Issue #7, opened 2018-11-28
Add strict scope mode and scope inheritance in non strict mode


klja:
Requested token should be validated against scopes defined on client.
If no scopes are requested they should be inherited from client scopes.

Proposed behavior for each use case:

| Storage | Client | Request | Token               |
|---------|--------|---------|---------------------|
| A, B, C | A      |         | A / Invalid scope   |
| A, B, C | A      | A       | A                   |
| A, B, C | A      | B       | Invalid scope       |
| A, B, C | A      | F       | Invalid scope       |
| A, B, C | F      |  *      | Invalid scope       |
| A, B, C |        | A       | A / Invalid scope   |
|         |        | *       | Invalid scope       |
| A,B, C  | A, B   |         | A,B / Invalid scope |

For examples that are defined with / (A / Invalid scope) - exact behavior would be defined through configuration. If _strict_ scope behaviour is set to _true_ Invalid scope exception is thrown, otherwise scope is inherited from client / storage.

spideyfusion:
Scope management has been revamped in https://github.com/trikoder/oauth2-bundle/pull/24.

PR #8, opened 2018-12-10
Add strict scope mode

klja:
Scopes missing in token request throw invalid token exception in strict mode.
In non strict mode, if scopes are missing in the request they are inherited from client or configuration #7


spideyfusion:
Scope management has been revamped in https://github.com/trikoder/oauth2-bundle/pull/24.

PR #10, opened 2018-12-19
Pass CryptKey instance to oauth2 with permission check disabled

gschafra:
Fix permission check (introduced with thephpleague/oauth2-server#776) for private keys
always failing in windows environments (also docker). Shamelessly stolen from laravel/passport#454.

### DIFF
--- a/DependencyInjection/TrikoderOAuth2Extension.php
+++ b/DependencyInjection/TrikoderOAuth2Extension.php
@@ -165,7 +165,11 @@ final class TrikoderOAuth2Extension extends Extension implements PrependExtensio
     {
         $container
             ->getDefinition('league.oauth2.server.resource_server')
-            ->replaceArgument('$publicKey', $config['public_key'])
+            ->replaceArgument('$publicKey', new Definition(CryptKey::class, [
+                $config['public_key'],
+                null,
+                false,
+            ]))
         ;
     }
 
--- a/DependencyInjection/TrikoderOAuth2Extension.php
+++ b/DependencyInjection/TrikoderOAuth2Extension.php
@@ -65,7 +65,11 @@ final class TrikoderOAuth2Extension extends Extension implements PrependExtensio
     {
         $authorizationServer = $container
             ->getDefinition('league.oauth2.server.authorization_server')
-            ->replaceArgument('$privateKey', new CryptKey($config['private_key'], null, false))
+            ->replaceArgument('$privateKey', new Definition(CryptKey::class, [
+                $config['private_key'],
+                null,
+                false,
+            ]))
             ->replaceArgument('$encryptionKey', $config['encryption_key'])
         ;
 
--- a/DependencyInjection/TrikoderOAuth2Extension.php
+++ b/DependencyInjection/TrikoderOAuth2Extension.php
@@ -12,6 +12,7 @@ use Symfony\Component\DependencyInjection\Extension\Extension;
 use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
+use League\OAuth2\Server\CryptKey;
 use Trikoder\Bundle\OAuth2Bundle\DBAL\Type\Grant as GrantType;
 use Trikoder\Bundle\OAuth2Bundle\DBAL\Type\RedirectUri as RedirectUriType;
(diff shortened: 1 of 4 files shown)


gschafra:
O.k... it's my first PR ever... so be tolerant :worried:

DependencyInjection/TrikoderOAuth2Extension.php
```diff
@@ -64,7 +65,7 @@ private function configureAuthorizationServer(ContainerBuilder $container, array
     {
         $authorizationServer = $container
             ->getDefinition('league.oauth2.server.authorization_server')
-            ->replaceArgument('$privateKey', $config['private_key'])
+            ->replaceArgument('$privateKey', new CryptKey($config['private_key'], null, false))
```
spideyfusion: You were on the right track, however, when you're altering the container you should use the `Definition` and `Reference` classes to describe the changes you're applying.

```suggestion
            ->replaceArgument('$privateKey', new Definition(CryptKey::class, [
                $config['private_key'],
                null,
                false,
            ]))
```

This is why Symfony wasn't able to resolve the environment variable placeholders correctly, as seen from the failed build output:

```
LogicException: Key path "file://%env(PRIVATE_KEY_PATH)%" does not exist or is not readable
```

spideyfusion: Also, you forgot to apply the changes for the public key path configuration in the `configureResourceServer` method.

spideyfusion: Since we're not checking for file permissions anymore, the `chmod` workaround from `Tests/TestKernel.php` in the `boot` method can be removed.

spideyfusion: We should also accompany the changes in tests as well. Check out the following places:

* `createAuthorizationServer` in `Tests/Integration/AbstractIntegrationTest.php`
* `createResourceServer` in `Tests/Integration/AbstractIntegrationTest.php`
* `generateJwtToken` in `Tests/TestHelper.php`

spideyfusion:
@gschafra Don't forget to [lint](https://github.com/trikoder/oauth2-bundle#code-linting) your code after you're done with the PR.

gschafra:
Unfortunately, when I'm trying to run  php via docker on my Windows machine here (company conventions, sorry), running Docker for Windows using:
```docker-compose run --rm --no-deps php -v``` 
i've got 
```shell
D:\vlinux\oauth2-bundle (disable-private-key-permission-check -> origin) λ docker-compose run --rm --no-deps php -v
WARNING: The HOST_USER_ID variable is not set. Defaulting to a blank string.
WARNING: The HOST_GROUP_ID variable is not set. Defaulting to a blank string.
sh: can't open 'docker-entrypoint': No such file or directory
```
Already tried to set ```autocrlf``` to ```input``` on cloning as suggested here: bigchaindb/bigchaindb#2215.
Any suggestions/ideas?

DependencyInjection/TrikoderOAuth2Extension.php
gschafra: Are you sure to remove the ```boot``` method at all, since it sets environment variables there?
```php   
    public function boot()
    {
        putenv(sprintf('PRIVATE_KEY_PATH=%s', TestHelper::PRIVATE_KEY_PATH));
        putenv(sprintf('PUBLIC_KEY_PATH=%s', TestHelper::PUBLIC_KEY_PATH));
        putenv(sprintf('ENCRYPTION_KEY=%s', TestHelper::ENCRYPTION_KEY));
    }
```
I'll leave it there for the moment, o.k.?

spideyfusion: Yeah, I only meant to remove the `chmod` calls so it looks like this:

```php
/**
 * {@inheritdoc}
 */
public function boot()
{
    putenv(sprintf('PRIVATE_KEY_PATH=%s', TestHelper::PRIVATE_KEY_PATH));
    putenv(sprintf('PUBLIC_KEY_PATH=%s', TestHelper::PUBLIC_KEY_PATH));
    putenv(sprintf('ENCRYPTION_KEY=%s', TestHelper::ENCRYPTION_KEY));

    parent::boot();
}
```

gschafra: Done

gschafra: Also done... sorry, dropped ```parent::boot()``` accidentally in comment. Thanks a lot for your help and input!

PR #17, opened 2019-01-17
Add console commands for client management

### LINKED ISSUE #3: Add console commands for client management

ajgarlag:
Fix #3

### DIFF
--- a/Command/CreateClientCommand.php
+++ b/Command/CreateClientCommand.php
@@ -81,7 +81,7 @@ final class CreateClientCommand extends Command
         return 0;
     }
 
-    protected function buildClientFromInput(InputInterface $input)
+    private function buildClientFromInput(InputInterface $input)
     {
         $identifier = $input->getArgument('identifier') ?? hash('md5', random_bytes(16));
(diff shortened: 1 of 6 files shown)


Command/CreateClientCommand.php
```diff
@@ -0,0 +1,112 @@
...
+        $headers = ['Identifier', 'Secret'];
+        $rows = [
+            [$client->getIdentifier(), $client->getSecret()],
+        ];
+        $io->table($headers, $rows);
+
+        return 0;
+    }
+
+    protected function buildClientFromInput(InputInterface $input)
```
alenpokos: @ajgarlag This method could be privated as it is only used in this command.

alenpokos:
@ajgarlag It would be nice to also have commands for listing and removing clients. Are you willing to PR those as well? You can do this in another PR or this one, it's up to you. From our perspective we can do it in two PRs.

ajgarlag:
@alenpokos I'd prefer to have this PR merged, and leave listing and removing commands for another one.

PR #18, opened 2019-01-17
Add support for the "authorization_code" grant type

### LINKED ISSUE #2: Add support for the "authorization_code" grant type

ajgarlag:
Fix #2


ajgarlag:
@spideyfusion Could you review this PR?What do you think about the new event to resolve the authorization decision?

Thanks in advance!

spideyfusion:
Great work thus far @ajgarlag! I definitely like the idea of having an event available where the developer has the chance to take complete control of the authorization process.

Could you give me a code example on how would you implement a simple decision page and an event listener which would appropriately approve or deny an authorization request?

ajgarlag:
I've written a simple proof of concept (not ready for production) which uses the `UriSigner` class from the `HttpKernel` component.

https://gist.github.com/ajgarlag/1f84d29ee0e1a92c8878f44a902338cd

MichaelKubovic:
Your implementation of authorization controller expects user to be already authenticated. The component demo says that you should first call `validateAuthorizationRequest()` and just then attempt authentication.

https://oauth2.thephpleague.com/authorization-server/auth-code-grant/

This allows SSO implementations where client (relying party in oidc terms) sends guests to authorization endpoint. The authorization endpoint validates request (authenticate against client). Then it should either re-use existing session or let user authenticate (could be redirect to route with form_login or any other configured firewall). The authentication on it's own is not in scope of OAuth, but we should be able to configure it.

Once logged in, user should be taken back to authorization endpoint to optionally authorize the client (this is where the event kicks in). Finally, user is redirected back with the authorization code and the flow continues on the RP side....

We could fix this easily with following changes:
- configure option to pickup previous session (check `IS_AUTHENTICATED_REMEMBERED` or stricter `IS_AUTHENTICATED_FULLY`)
- optionally configure login route
- store validated request in application session, redirect if login url configured, else throw unauthorized exception
- restore oauth authorization, either directly by appending callback query param to login url or by providing authentication listener that one has to use to pickup the flow and continue.

ajgarlag:
Good catch, but I prefer a different way to fix it.

That check with the AuthorizationChecker is there from the first commit, before the `AuthorizationRequestResolveEvent` was introduced.

Now, I would move the responsibility of checking the user authentication and the responsibility of setting the UserEntity into the event, to two different event listeners. The first one will set the attribute required to process the `AuthorizationRequest`. I would put it to `IS_AUTHENTICATED_REMEMBERED` by default, but it could be overridden by configuration.

This will allow to prepend an EventListener from an OIDC implementation that will manage the `prompt` parameter.

What do you think?

MichaelKubovic:
Sounds good to me. As long as we are able to plug-in any authentication provided by Symfony Security, I'd be happy :)

ajgarlag:
Finally, I have decided to remove any authorization check from the authorization endpoint. It's not needed at all. It is up to you to decide how to protect the authorization endpoint.

Anyway, I have added a recommendation to the README to protect it, so that anonymous users cannot access to the authorization endpoint and cannot approve authorization requests.

MichaelKubovic:
I believe this is not entirely correct. Anonymous users are welcome to authorization endpoint. See https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowSteps
It's just after you validated authorization request that you **authenticate** user. What I miss is redirect to login page if TokenStorage does not provide authenticated token.

ajgarlag:
Yes, I know that, but remember that we are implementing here the authorization_code grant for an oAuth 2.0 server, not for an OIDC provider.

In an OIDC provider implementation, the security configuration should allow anonymous access to authorization endpoint, and an EventListener will have to manage the `prompt` parameter, and redirect to the login if needed; but in this PR we does not need anything like that.

Anyway, the access to authorization endpoint it's not limited by default. If you know what are you doing, you can omit the recommendation and allow anonymous users access.

MichaelKubovic:
👍 @spideyfusion what do you think, ready to merge?

ajgarlag:
@spideyfusion The PR is ready to merge for me.

MichaelKubovic:
I've been working with the code in this PR and I dislik how the AuthorizationRequestResolve event in AuthorizationController is tightly coupled to the listener that reacts to it. I've tried to make it a bit more flexible, so I've moved the responsibility to create response or approve authorization request to listener.

Furthermore I've encapsulated the authorization approval bit so that we can employ any other strategy.

I've created a PR against @ajgarlag 's feature branch so you could review it separately
https://github.com/ajgarlag/oauth2-bundle/pull/1/files

Resources/config/routes.xml
```diff
@@ -3,5 +3,6 @@
         xsi:schemaLocation="http://symfony.com/schema/routing
         http://symfony.com/schema/routing/routing-1.0.xsd">
 
+    <route id="oauth2_authorize" path="/authorize" controller="trikoder.oauth2.controller.authorization_controller:indexAction" methods="GET|POST" />
```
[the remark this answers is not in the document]
ajgarlag: From [oAuth2 RFC](https://tools.ietf.org/html/rfc6749#section-3.1):

> The authorization server MUST support the use of the HTTP "GET"
> method [RFC2616] for the authorization endpoint and MAY support the
> use of the "POST" method as well.

From [OpenID Connect Core 1.0 spec](https://openid.net/specs/openid-connect-core-1_0.html#rfc.section.3.1.2.1):
> Authorization Servers MUST support the use of the HTTP GET and POST methods...

Resources/config/doctrine/model/AuthCode.orm.xml
```diff
@@ -0,0 +1,19 @@
+<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
```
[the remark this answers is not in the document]
ajgarlag: I chose the AuthCode name to follow the [league/oauth2-server](https://github.com/thephpleague/oauth2-server) naming, but if you prefer it, I can rename AuthCode to AuthorizationCode. @alenpokos, can you confirm it?

Tests/Integration/AuthorizationServerTest.php
```diff
@@ -383,4 +383,206 @@ public function testInvalidPayloadRefreshGrant(): void
...
+        $request = $this->createAuthorizeRequest(null, [
+            'response_type' => 'code',
+            'client_id' => 'foo',
+            'redirect_uri' => 'https://example.org/oauth2/redirect-uri',
+        ]);
+
+        $response = $this->handleAuthorizeRequest($request);
+
+        // Response assertions.
+        $this->assertArrayHasKey('code', $response);
```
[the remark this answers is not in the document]
ajgarlag: @alenpokos Do you think that `AbstractIntegrationTest::handleAuthorizeRequest` should return a `ResponseInterface` instead of an array?

I tried to mimic the `AbstractIntegrationTest::handleAuthorizationRequest` behavior.

Tests/Acceptance/AuthorizationEndpointTest.php
```diff
@@ -0,0 +1,98 @@
...
+
+        timecop_return();
+
+        $response = $this->client->getResponse();
+
+        $this->assertSame(302, $response->getStatusCode());
+        $redirectUri = $response->headers->get('Location');
+        $this->assertEquals('/authorize/consent', $redirectUri);
+    }
+
```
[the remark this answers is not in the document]
ajgarlag: Done in b6f385aa6f15a7f60a980aa376fc5d4e2ab27bdd

ajgarlag:
@MichaelKubovic Thanks for your PR, I'll review it after all requested changes by @alenpokos are addressed.

Tests/Integration/AbstractIntegrationTest.php
```diff
@@ -186,12 +211,39 @@ protected function handleResourceRequest(ServerRequestInterface $serverRequest):
         return $serverRequest;
     }
 
+    protected function handleAuthorizeRequest(ServerRequestInterface $serverRequest, $approved = true): array
```
alenpokos: @ajgarlag Looking at method name now, it seems unclear compared to handleAuthorizationRequest. I would suggest renaming this method to ie. `handleAuthorizationCodeGrantRequest` and @spideyfusion or I can rename `handleAuthorizationRequest` to more clear name as well. What do you think?

Tests/Integration/AuthorizationServerTest.php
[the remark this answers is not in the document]
alenpokos: @ajgarlag Agree, returning ResponseInterface would be more sensible and test can verify what is a result of request.

Resources/config/routes.xml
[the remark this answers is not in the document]
alenpokos: @ajgarlag Valid point, we can have support both.

Resources/config/doctrine/model/AuthCode.orm.xml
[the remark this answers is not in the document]
alenpokos: @ajgarlag I think AuthorizationCode would be better.

Tests/Integration/AbstractIntegrationTest.php
ajgarlag: This method will be used to test implicit grant too, so the `handleAuthorizationCodeGrantRequest` name is not very good.

My first option was to rename `handleAuthorizationRequest` to `handleTokenRequest`, and name the new method `handleAuthorizationRequest` but I discarded it to avoid a BC break. But if you think that this method can be renamed, this will be my preferred option. @alenpokos What do you think?

ajgarlag: @alenpokos Can you confirm me if renaming the `handleAuthorizationRequest` is an acceptable solution? Are you going to rename it in another PR, or do you prefer to include the change in this one?

alenpokos: @ajgarlag Sry for wait. Yes, the name sounds good and I think it is ok to rename it in this MR as this is where we noticed the need to change.

ajgarlag:
@alenpokos I've introduced some changes in the the AuthorizationRequestResolveEvent workflow.

The `authorizationResolution` is a boolean value, `false` by default. Instead of a `resolutionUri` the event can have a `ResponseInterface`.

The controller checks if the event has a `ResponseInterface`. If it has one, the controller should return that response. If the event has no response, the controller should resolve the authorization request with the `authorizationResolution` value.

When `AuthorizationRequestResolveEvent::resolveAuthorization` is called, the event discards the `ResponseInterface` if it exists.

Controller/AuthorizationController.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
```
Derkien: @ajgarlag Repository is about PHP7.2+. It would be great if you provide ```declare(strict_types=1)``` to every class and provide type for every argument in methods (sometimes it's missed). It seems it must be done for all project.

ajgarlag: @Derkien You are right, it must be done for all files in project, so I think it should be done in a different PR modifying the `.php_cs.dist` file to accomplish it. WDYT?

spideyfusion: @ajgarlag @Derkien Definitely in a separate PR so we don't go out of scope in this one.

Derkien: @ajgarlag Sounds like a plan :) 👍 
@spideyfusion Ofc in separate.

ajgarlag:
@alenpokos @spideyfusion Is there any pending change?

MichaelKubovic:
ping, what's the state of this issue?

froozeify:
Any news on this pr ?

spideyfusion:
We'll take a look at resolving this PR by the end of the week. Sorry for the lack of communication, we've been extremely busy this past month.

Controller/AuthorizationController.php
```diff
@@ -0,0 +1,78 @@
...
+    public function __construct(AuthorizationServer $server, Security $security, EventDispatcherInterface $eventDispatcher)
+    {
+        $this->server = $server;
+        $this->security = $security;
+        $this->eventDispatcher = $eventDispatcher;
+    }
+
+    public function indexAction(ServerRequestInterface $serverRequest): ResponseInterface
+    {
+        $serverResponse = new Response();
```
X-Coder264: In https://github.com/trikoder/oauth2-bundle/pull/43 we made the effort to decouple the bundle from the `Zend\Diactoros` concrete PSR-7 implementation. Please use the `Psr\Http\Message\ResponseFactoryInterface` to create the response object so that the users of the bundle can use whatever PSR-7 implementation they like.

Event/AuthorizationRequestResolveEvent.php
```diff
@@ -0,0 +1,115 @@
...
+     * @var bool
+     */
+    private $authorizationResolution = self::AUTHORIZATION_DENIED;
+
+    public function __construct(AuthorizationRequest $authorizationRequest)
+    {
+        $this->authorizationRequest = $authorizationRequest;
+    }
+
+    public function getAuhorizationResolution(): bool
```
X-Coder264: You have a typo here: `getAuhorizationResolution` -> `getAuthorizationResolution` 😄

X-Coder264:
The PR looks good, good job @ajgarlag 👍 Please rebase with master (some conflicts exist now in a couple of tests) and check the two small comments that I've left so that we can merge this.

Event/AuthorizationRequestResolveEvent.php
```diff
@@ -0,0 +1,115 @@
...
+    public const AUTHORIZATION_APPROVED = true;
+    public const AUTHORIZATION_DENIED = false;
+
+    /**
+     * @var AuthorizationRequest
+     */
+    private $authorizationRequest;
+
+    /**
+     * @var ?ResponseInterface
```
X-Coder264: `@var ResponseInterface|null` so that it's consistent with the rest of the bundle.

ajgarlag:
> The PR looks good, good job @ajgarlag +1 Please rebase with master (some conflicts exist now in a couple of tests) and check the few small comments that I've left so that we can merge this.

@X-Coder264 Thanks for the review. I'll squash all commits before rebase and will check the comments.

ajgarlag:
There are at least two changes proposed by @MichaelKubovic that I think we could add to this PR:

1. Delegate the user resolution to an event listener. Current implementation avoid any mutation to the AuthorizationRequest object, but I think one makes sense. See https://github.com/ajgarlag/oauth2-bundle/commit/a74a540e6170d4dea60ddddeee4c6d3898d9c180
2. The event listener with higher priority should win. Current implementation allows any event listener to modify the decision taken by an already executed listener. We can change this, so no event listener is executed if its decision is going to be discarded. If you want to override the decision of an event listener, just add another one with higher priority. See https://github.com/ajgarlag/oauth2-bundle/commit/1a8edbd4de8edb10d8aa72fb1d329f50352ec270

Event/AuthorizationRequestResolveEvent.php
```diff
@@ -0,0 +1,115 @@
+<?php
+
+namespace Trikoder\Bundle\OAuth2Bundle\Event;
+
+use League\OAuth2\Server\Entities\ClientEntityInterface;
```
spideyfusion: @ajgarlag We should make the return values consistent with existing event objects. In other words, return bundle specific entities instead of the ones provided by League. This also includes targeting `UserInterface` provided by Symfony.

Examples:

* https://github.com/trikoder/oauth2-bundle/blob/master/Event/ScopeResolveEvent.php#L60
* https://github.com/trikoder/oauth2-bundle/blob/master/Event/UserResolveEvent.php#L70

spideyfusion:
@ajgarlag It makes sense to integrate those two commits you linked. We can merge this in after you cherry-pick them and address my final feedback. :)

Next up, we'll focus on writing documentation in a separate PR and then we'll tag a new major release.

Event/AuthorizationRequestResolveEvent.php
ajgarlag: Not sure about this. To address this, I should change at least 4 methods:
```php
final class AuthorizationRequestResolveEvent extends Event
{
//...
    public function getClient(): ClientEntityInterface
    {
        return $this->authorizationRequest->getClient();
    }

    public function getUser(): UserEntityInterface
    {
        return $this->authorizationRequest->getUser();
    }

    public function setUser(UserEntityInterface $user): void
    {
        $this->authorizationRequest->setUser($user);
    }

    /**
     * @return ScopeEntityInterface[]
     */
    public function getScopes(): array
    {
        return $this->authorizationRequest->getScopes();
    }

/...
}
```
* The `getScopes` method could use the `ScopeConverter` helper to transform league entities into bundle models, so no problem here.
* The `setUser` method could receive a Symfony `UserInterface` instance an transform it into a League `UserEntityInterface` before setting it into the `AuthorizationRequest` object.
* With the `getUser` method I do not see how to transform the `UserEntityInterface` from the `AuthorizationRequest` into a Symfony `UserInterface`. We could store the instance received from the `setUser` method in a private property, but if the `AuthorizationRequest` received as constructor parameter already has an user set, how can we transform it into an Syfmony `UserInterface` to expose this info to the listeners?
* Finally, in the `getClient` method, we need to have a reference to a `ClientManagerInterface` instance to find the `Model\Client` using the `ClientEntityInterface` identifier.

WDYT?

Event/AuthorizationRequestResolveEvent.php
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    public function getUser(): UserEntityInterface
+    {
+        return $this->authorizationRequest->getUser();
+    }
+
+    public function setUser(UserEntityInterface $user): void
+    {
+        $this->authorizationRequest->setUser($user);
```
ajgarlag: @spideyfusion Should we check if the `AuthorizationRequest` already has an user set, and throw a `LogicException`,  if a listener tries to overwrite it?

ajgarlag:
> Next up, we'll focus on writing documentation in a separate PR and then we'll tag a new major release.

@spideyfusion before tagging a new major release, I'd like to include #21. I'll rebase it once #18 gets merged.

Event/AuthorizationRequestResolveEvent.php
spideyfusion: @ajgarlag I took a look at League's implementation and they don't seem to set the user themselves anywhere in the library. This is the only place where the `AuthorizationRequest` gets instantiated:

https://github.com/thephpleague/oauth2-server/blob/7.4.0/src/Grant/AuthCodeGrant.php#L272

With that being said, I think we can still remain consistent regarding `UserInterface`. Here are my suggestions:

* Change the `AuthorizationRequestResolveEvent` event so it doesn't directly manipulate the `AuthorizationRequest` object. Events should only communicate the desired state. We can set the user on `AuthorizationRequest` from where the event gets dispatched (the controller).
* If multiple services are required to construct the `AuthorizationRequestResolveEvent`, we can move it to a dedicated class.
* It seems we have a need to convert the entities yet again. Perhaps it's time to introduce a class similar to the `ScopeConverter` so we avoid code duplication. Here's what I'm [referring to](https://github.com/trikoder/oauth2-bundle/blob/master/League/Repository/UserRepository.php#L61).

These changes would also make `AuthorizationRequestUserResolvingListener` way simpler.

ajgarlag: >  * Change the AuthorizationRequestResolveEvent event so it doesn't directly manipulate the AuthorizationRequest object. Events should only communicate the desired state. We can set the user on AuthorizationRequest from where the event gets dispatched (the controller).

Done.
>  * If multiple services are required to construct the AuthorizationRequestResolveEvent, we can move it to a dedicated class.

Done, but I'm very bad at naming things, so maybe the event factory should be renamed and/or relocated.
>  * It seems we have a need to convert the entities yet again. Perhaps it's time to introduce a class similar to the ScopeConverter so we avoid code duplication. Here's what I'm referring to

Done.

OAuth2Events.php
```diff
@@ -21,4 +21,13 @@ final class OAuth2Events
      * You could alter the access token's scope here.
      */
     public const SCOPE_RESOLVE = 'trikoder.oauth2.scope_resolve';
+
+    /**
+     * The AUTHORIZATION_REQUEST_RESOLVE event occurrs right before the system
+     * complete the authorization request.
+     *
+     * You could approve or deny the authorization request, or set the uri where
```
Nyholm: > "or set the uri"

You mean or add a `RedirectResponse` to the event?

Question: If I redirect the user to a form where they accept or deny the claim. How do I get back? 
Could you add some more documentation about this?

(Im currently testing this PR)
Thank you for you work!

ajgarlag: Yes, I mean "or set the response, (usually a redirect response), that will be returned to the user to resolve the authorization process"

Regarding your question, you can have a listener which can extract the user decision from the request. See https://github.com/trikoder/oauth2-bundle/pull/18#issuecomment-460157339

Event/AuthorizationRequestResolveEvent.php
spideyfusion: @ajgarlag Looking good! 👍

spideyfusion:
Great work @ajgarlag! Next up, let's integrate https://github.com/trikoder/oauth2-bundle/pull/21 into the `master` branch.

DonCallisto:
From what I see here, this bundle still misses the part of `Authentication` (that's not the scope of OAuth2.0 I know). I mean, even with the snippet of code provided above, you are implicitly saying that the `Authorization Provider/Endpoint/Server` (call it whatever you want) also is the reponsible for Authenticating the user, for displaying a consent dialog to him and for having some resources to provide.
That's quite troublesome from my POV as AuthenticationProvider *could* be even a different service, without any knowledge of the users as its main reason to exist is to provide token and verify that tose tokens are valid (whatever "valid" means here).
So basically, just to make an example, I can't provide a microservice responsible only for tokens and a different resource endpoint.
Am I right?

DonCallisto:
Well, so this means that also `AuthorizationProvider` has authenticated user (as long as them should be into `TokenStorage`. Moreover the `Listener` is something "syncronous", what if (and that's my case) I need the `GET ..../authorization` endpoint to be redirected to `LoginProvider` (call it this way, I don't have any better name as is not part of the protocol) and then to the `ConsentProvider` (ok, it could be the same of `LoginProvider` but just to give it different names)

MichaelKubovic:
What I've done is, that I accept anonymous users /authorization on endpoint. I registered another event listener, that runs before the `AuthorizationRequestUserResolvingListener ` and redirects appropiately. https://github.com/MichaelKubovic/oauth2-bundle/blob/develop/EventListener/AuthorizationRequestAuthenticationResolvingListener.php

You can take user to 3rd party service if you wish.

DonCallisto:
@MichaelKubovic well, so how do you "restore" the user from the redirect? You need to have access to *same* db and you need to return with some information from the 3rd party (like an id).
And what about user consent? Is provided by the 3rd party *after* de login? And how do you return from the 3rd party login to the "original" `/authorization` call if is handled by the 3rd party?

MichaelKubovic:
I do authentication within the same Symfony instance. But this bundle does not force you. For the consent flow, I have yet another listener hooked to the same authorization event.

It's not the responsibility of this bundle, neither the responsibility of oauth 2 to deal with details of authentication.

DonCallisto:
> It's not the responsibility of this bundle, neither the responsibility of oauth 2 to deal with details of authentication.

I know this but as long as you need a logged user *inside* this bundle I suppose that you're assuming that everything is tied togheter: `AuthorizationRequest` needs a `UserEntityInterface` in order to work, it should rely on some other thing like an identifier or something.

The fact that you're doint authentication inside the same symfony instance maybe don't let you grasp the problem or probably I'm explaining it in the wrong way.

DonCallisto:
Take a look at [this](https://www.ory.sh/docs/hydra/oauth2#authenticating-users-and-requesting-consent) especially when it says

> As you already know by now, ORY Hydra does not come with any type of user management (login, registration, ...).

That's the key aspect as `User` concept is not a OAuth concept.

MichaelKubovic:
I see what confuses you. The interface that this bundle relies on, the `Symfony\Component\Security\Core\User\UserInterface`, is often used to implement user in the context of Symfony Security and it's related concepts. It's just convenient. It's a Symfony bundle after all. It does not imply authentication by security component. It implies instance of user must be provided by you.

Think of `UserEntityInterface` in your symfony service as a resolved identity. The object does not have to come from the database. It does not have to be used to really authenticate user. It just must exist in token storage for this bundle to work.

If you have authentication implemented elsewhere, you should know how to communicate with it. Do you federate to OIDC provider? Do you share cookie? Do you use some means of one-time access token? Up to you.

> That's the key aspect as User concept is not a OAuth concept.
Disagree, the User is mentioned in spec many times. See https://tools.ietf.org/html/rfc6749#section-4.1. Spec says that the user should be authenticated through user agent.

DonCallisto:
I know the key concept of symfony, I just only mistaken the interface that's not from Symfony but from league bundle (that has also the `getPassword` method, meaning that you need some kind of authentication on it).
Nevermind, I get your point.

PR #24, opened 2019-02-04
Implement client scope inheritance and restriction

spideyfusion:


League/Repository/ScopeRepository.php
```diff
@@ -83,4 +88,32 @@ public function finalizeScopes(
 
         return $this->scopeConverter->toLeagueArray($event->getScopes());
     }
+
+    /**
+     * @return ScopeModel[]
+     */
+    private function setupScopes(ClientModel $client, array $requestedScopes): array
```
alenpokos: @spideyfusion I would suggest to change return type to be iterable

Tests/Fixtures/FixtureFactory.php
```diff
@@ -13,6 +13,12 @@
 use Trikoder\Bundle\OAuth2Bundle\Model\RefreshToken;
 use Trikoder\Bundle\OAuth2Bundle\Model\Scope;
 
+/**
+ * Development hints:
```
alenpokos: :+1:

Tests/Integration/AuthorizationServerTest.php
```diff
@@ -103,6 +103,21 @@ public function testRestrictedGrantClient(): void
         $this->assertSame('Client authentication failed', $response['message']);
     }
 
+    public function testRestrictedScopeClient(): void
+    {
+        $request = $this->createAuthorizationRequest('quux_restricted_scopes:beer', [
+            'grant_type' => 'client_credentials',
+            'scope' => 'fancy rock',
```
alenpokos: @spideyfusion Can you please use fixture constants instead of concrete values? Now it is mixed between two and makes it hard to track.

Tests/Integration/AuthorizationServerTest.php
```diff
@@ -187,6 +202,37 @@ public function testValidClientCredentialsGrantWithScope(): void
         );
     }
 
+    public function testValidClientCredentialsGrantWithInheritedScope(): void
+    {
+        $request = $this->createAuthorizationRequest('quux_restricted_scopes:beer', [
```
alenpokos: @spideyfusion Can we also use client IDs from constants? This makes it a lot easier to follow the test

Event/ScopeResolveEvent.php
```diff
@@ -45,9 +45,6 @@ public function getScopes(): array
         return $this->scopes;
     }
 
-    /**
-     * @param Scope[] $scopes
-     */
     public function setScopes(Scope ...$scopes): self
     {
         $this->scopes = (array) $scopes;
```
Derkien: This type casting is not necessary, as the argument is of needed type.

League/Repository/ScopeRepository.php
Derkien: Why? Both methods: ClientModel->getScopes, scopeConverter->toDomainArray and current method logic provide clean return type "array"

spideyfusion: @Derkien You bring up a good point. I have reverted the change.

Event/ScopeResolveEvent.php
spideyfusion: @Derkien Good catch.

Tests/Integration/AuthorizationServerTest.php
spideyfusion: @alenpokos This will be resolved in a separate PR where we'll tackle test readability.

Tests/Integration/AuthorizationServerTest.php
spideyfusion: @alenpokos This will be resolved in a separate PR where we'll tackle test readability.

PR #51, opened 2019-05-18
Add declare strict types CS fixer rule

X-Coder264:


Tests/Integration/AbstractIntegrationTest.php
```diff
@@ -171,7 +178,7 @@ protected function handleAuthorizationRequest(ServerRequestInterface $serverRequ
             $response = $e->generateHttpResponse($response);
         }
 
-        return json_decode($response->getBody(), true);
+        return json_decode($response->getBody()->__toString(), true);
```
HypeMC: @X-Coder264 Why not cast?

Tests/Integration/AbstractIntegrationTest.php
```diff
@@ -144,12 +146,17 @@ protected function getRefreshToken(string $encryptedPayload): ?RefreshToken
...
-        return $this
+        $request = $this
             ->psrFactory
             ->createServerRequest('', '')
-            ->withHeader('Authorization', sprintf('Basic %s', base64_encode($credentials)))
             ->withParsedBody($body)
         ;
+
+        if (null !== $credentials) {
+            return $request->withHeader('Authorization', sprintf('Basic %s', base64_encode($credentials)));
```
HypeMC: @X-Coder264 Do we need this return here? Don't see any real benefit, what if more conditions are added in the future?

X-Coder264: @HypeMC We need it, because the PSR-7 request is immutable and a new request instance is returned on the `withHeader` call.

Tests/Integration/AbstractIntegrationTest.php
X-Coder264: @HypeMC I used it because  the `__toString` method is actually present on the `StreamInterface` PSR-7 interface.

Tests/Integration/AbstractIntegrationTest.php
HypeMC: @X-Coder264 Fair enough, but still what if more conditions are added later, what about `$request = $request->withHeader`? Your call, I'm just writing my thoughts.

X-Coder264: @HypeMC Good point, done.

PR #52, opened 2019-05-18
Make the enabled grant types configurable

### LINKED ISSUE #34: Should the enabled grant types be configurable?
Hello,

I am currently in the process of deciding to use this bundle or directly the underlying library from the PHP league.

The issues that I have are:
- I would need to configure which grant types are enabled. I do not think it is possible a the moment since [they are enabled in the bundle extension](https://github.com/trikoder/oauth2-bundle/blob/f16ec67f2fa8dbf8fedd78488d625cef2db5b90d/DependencyInjection/TrikoderOAuth2Extension.php#L77-L90).
- I would need to override the password grant type with a custom one since the provided password grant is not flexible enough for us: we would need more than username and password in order to authenticate users. At the moment, the only way I can imagine to do that is to override the service `league.oauth2.server.grant.password_grant` and hope it still works when the bundle gets updated.

1. Do you think that it would be a good idea to make the enabled grant types configurable?
2. Would you consider a pull request about this?

Kind regards,

Guillaume.

X-Coder264:
The grants are enabled by default to preserve backwards compatibility.

Closes #34

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -59,6 +59,18 @@ final class Configuration implements ConfigurationInterface
                     ->cannotBeEmpty()
                     ->defaultValue('P1M')
                 ->end()
+                ->booleanNode('enable_client_credentials_grant')
+                    ->info('Whether to enable the client credentials grant')
+                    ->defaultTrue()
+                ->end()
+                ->booleanNode('enable_password_grant')
+                    ->info('Whether to enable the password grant')
+                    ->defaultTrue()
+                ->end()
+                ->booleanNode('enable_refresh_token_grant')
+                    ->info('Whether to enable the refresh token grant')
+                    ->defaultTrue()
+                ->end()
             ->end()
         ;
 
--- a/Tests/Unit/ExtensionTest.php
+++ b/Tests/Unit/ExtensionTest.php
@@ -1,0 +1,88 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Trikoder\Bundle\OAuth2Bundle\Tests\Unit;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Trikoder\Bundle\OAuth2Bundle\DependencyInjection\TrikoderOAuth2Extension;
+use Trikoder\Bundle\OAuth2Bundle\Manager\InMemory\ScopeManager;
+use Trikoder\Bundle\OAuth2Bundle\Manager\ScopeManagerInterface;
+
+final class ExtensionTest extends TestCase
+{
+    /**
+     * @dataProvider grantsProvider
+     */
+    public function testEnablingAndDisablingGrants(string $referenceId, string $grantKey, bool $shouldTheGrantBeEnabled): void
+    {
+        $container = new ContainerBuilder();
+
+        $this->setupContainer($container);
+
+        $extension = new TrikoderOAuth2Extension();
+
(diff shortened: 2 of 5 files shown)


Tests/Unit/ExtensionTest.php
```diff
@@ -0,0 +1,88 @@
...
+     */
+    public function testEnablingAndDisablingGrants(string $referenceId, string $grantKey, bool $shouldTheGrantBeEnabled): void
+    {
+        $container = new ContainerBuilder();
+
+        $this->setupContainer($container);
+
+        $extension = new TrikoderOAuth2Extension();
+
+        $extension->load($this->getValidConfiguration([$grantKey => $shouldTheGrantBeEnabled]), $container);
```
HypeMC: @X-Coder264 You could always do an `array_merge` instead of passing `$options`, but I don't know whether that's better or worse tbh.

X-Coder264: @HypeMC I was thinking of doing that originally, but this seems cleaner to me for now (as the array is more than one level deep).

Tests/Unit/ExtensionTest.php
```diff
@@ -0,0 +1,88 @@
...
+                break;
+            }
+        }
+
+        $this->assertSame($shouldTheGrantBeEnabled, $isGrantEnabled);
+    }
+
+    public function grantsProvider(): iterable
+    {
+        return [
```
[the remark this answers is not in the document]
X-Coder264: @HypeMC I guess it's just a matter of preference since in this case the array is quite small, but I've changed it nonetheless :)

DependencyInjection/Configuration.php
```diff
@@ -57,6 +57,18 @@ private function createAuthorizationServerNode(): NodeDefinition
                     ->cannotBeEmpty()
                     ->defaultValue('P1M')
                 ->end()
+                ->booleanNode('enable_client_credentials_grant')
+                    ->info('Whether to enable the client credentials grant')
+                    ->defaultTrue()
```
HypeMC: @X-Coder264 Do we want the default to be true? Since we'll be bumping the major version it wouldn't be a bc break.

X-Coder264: @HypeMC It's true that we could make the default to be disabled as this PR will land in the next major version, but tbh I don't see any benefit in creating a breaking change in this case. From my perspective this PR aims to give the users the option to disable some grant if they really need to do it for whatever reason. Having all of them enabled by default seems ok too me. What's your take on this?

HypeMC: @X-Coder264 The reason I'm asking is because it's very unlikely that someone will need more then one or two grant types, so it might be better to force the users to explicitly enable what they need instead of relying on them to know or remember to disable what they don't need.

X-Coder264: @HypeMC I don't like to force the users to do something that they really don't need to do. Having all grant types enabled seems (at least to me) like a sane default while having to disable a grant type seems like an edge case to me (and only then would the users have to check how to disable them, as opposed to the opposite scenario where each user would have to enable the grants they need as an additional step before they could begin to use the bundle). Laravel Passport also enables all grant types (except the implicit one) by default -> https://github.com/laravel/passport/blob/7.0/src/PassportServiceProvider.php#L95 That's just my 2c though.

HypeMC: @X-Coder264 Works for me, just wanted to here you opinion on the matter.

X-Coder264: @HypeMC 👍

PR #53, opened 2019-05-19
Add a console command for removing expired tokens

### LINKED ISSUE #4: Add console commands for stale token cleanup

HypeMC:
Closes #4

### DIFF
--- a/Command/ClearExpiredTokensCommand.php
+++ b/Command/ClearExpiredTokensCommand.php
@@ -1,0 +1,91 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Trikoder\Bundle\OAuth2Bundle\Command;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Trikoder\Bundle\OAuth2Bundle\Manager\AccessTokenManagerInterface;
+use Trikoder\Bundle\OAuth2Bundle\Manager\RefreshTokenManagerInterface;
+
+final class ClearExpiredTokensCommand extends Command
+{
+    protected static $defaultName = 'trikoder:oauth2:clear-expired-tokens';
+
+    /**
+     * @var AccessTokenManagerInterface
+     */
+    private $accessTokenManager;
+
+    /**
+     * @var RefreshTokenManagerInterface
+     */
+    private $refreshTokenManager;
+
+    public function __construct(
+        AccessTokenManagerInterface $accessTokenManager,
+        RefreshTokenManagerInterface $refreshTokenManager
+    ) {
+        parent::__construct();
+
+        $this->accessTokenManager = $accessTokenManager;
+        $this->refreshTokenManager = $refreshTokenManager;
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setDescription('Clears all expired access and/or refresh tokens')
+            ->addOption(
+                'access-tokens-only',
+                'a',
+                InputOption::VALUE_NONE,
+                'Clear only access tokens.'
+            )
+            ->addOption(
+                'refresh-tokens-only',
+                'r',
+                InputOption::VALUE_NONE,
+                'Clear only refresh tokens.'
+            )
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+
+        $clearExpiredAccessTokens = !$input->getOption('refresh-tokens-only');
+        $clearExpiredRefreshTokens = !$input->getOption('access-tokens-only');
+
+        if (!$clearExpiredAccessTokens && !$clearExpiredRefreshTokens) {
+            $io->error('Please choose only one of the following options: "access-tokens-only", "refresh-tokens-only".');
+
+            return 1;
+        }
+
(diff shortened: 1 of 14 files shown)


Command/ClearExpiredTokensCommand.php
```diff
@@ -0,0 +1,89 @@
...
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+
+        $clearExpiredAccessTokens = !$input->getOption('refresh-tokens-only');
+        $clearExpiredRefreshTokens = !$input->getOption('access-tokens-only');
+
+        if (!$clearExpiredAccessTokens && !$clearExpiredRefreshTokens) {
+            $io->error('Please chose only one of the following options: "access-tokens-only", "refresh-tokens-only".');
```
X-Coder264: @HypeMC Chose -> choose typo :)

Manager/Doctrine/AccessTokenManager.php
```diff
@@ -34,4 +34,13 @@ public function save(AccessToken $accessToken): void
         $this->entityManager->persist($accessToken);
         $this->entityManager->flush();
     }
+
+    public function clearExpired(): int
+    {
+        return $this->entityManager->createQueryBuilder()
+            ->delete(AccessToken::class, 'at')
+            ->where('at.expiry < CURRENT_TIMESTAMP()')
```
X-Coder264: @HypeMC I'm not sure how I feel about the usage of `CURRENT_TIMESTAMP()` here. Using it means that this feature won't work on all databases supported by Doctrine. On the other hand binding an object (DateTime in this case) via the `setParameter()` method is "broken" (https://github.com/doctrine/orm/pull/7528) in Doctrine 2.6.3 (in our internal application that uses this bundle) and 2.64 which has the fix for it hasn't been tagged yet. What is your opinion on this matter?

Tests/Acceptance/ClearExpiredTokensCommandTest.php
```diff
@@ -0,0 +1,108 @@
...
+        $commandTester = new CommandTester($command);
+
+        $exitCode = $commandTester->execute([
+            'command' => $command->getName(),
+        ]);
+
+        $this->assertSame(0, $exitCode);
+
+        $output = $commandTester->getDisplay();
+        $this->assertStringContainsString('Cleared 1 expired access token.', $output);
```
X-Coder264: @HypeMC These and some other similar tests in this PR don't actually test the feature in question. They only test the string output of the command which makes them kinda useless in its current state. The test should have assertions that prove that the tokens get actually cleared properly in the database.

Tests/Unit/InMemoryAccessTokenManagerTest.php
```diff
@@ -0,0 +1,66 @@
...
+        timecop_freeze(new DateTime());
+
+        $testData = $this->buildClearExpiredTestData();
+
+        foreach ($testData['input'] as $token) {
+            $inMemoryAccessTokenManager->save($token);
+        }
+        $this->assertSame(3, $inMemoryAccessTokenManager->clearExpired());
+
+        timecop_return();
```
X-Coder264: @HypeMC Please use the setUp/tearDown methods for (un)freezing time or wrap this chunk of the code in a try {} finally {} block. The way the test is currently written if the assertion above the `timecop_return` call fails an exception will be thrown which means that `timecop_return` won't be called and all the rest of the tests will be run with frozen time. The same applies for the `InMemoryRefreshTokenManagerTest` too.

docs/basic-setup.md
```diff
@@ -153,6 +153,24 @@ There are two possible reasons for the authentication server to reject a request
 - Provided token is expired or invalid (HTTP response 401 `Unauthorized`)
 - Provided token is valid but scopes are insufficient (HTTP response 403 `Forbidden`)
 
+## Clearing expired access & refresh tokens
+
+To clear expired access & refresh tokens you can use the `trikoder:oauth2:clear-expired-tokens` command.
+
+The command removes all tokes whose expiry time is lesser than the current.
```
X-Coder264: @HypeMC Tokes -> tokens typo :)

Manager/Doctrine/AccessTokenManager.php
X-Coder264: nvm, CURRENT_TIMESTAMP() is actually a DQL function which gets parsed properly.

Command/ClearExpiredTokensCommand.php
HypeMC: Done

docs/basic-setup.md
HypeMC: Done

Tests/Acceptance/ClearExpiredTokensCommandTest.php
HypeMC: @X-Coder264 Added two additional tests, currently in the acceptance tests folder, hopefully will be moved at a later time once the infrastructure is improved.

Manager/Doctrine/AccessTokenManager.php
HypeMC: @X-Coder264 Switched to DateTime after all for consistency & cause timecop doesn't work with `CURRENT_TIMESTAMP`.

Tests/Unit/InMemoryAccessTokenManagerTest.php
HypeMC: @X-Coder264 Agreed, there are several tests that use timecop incorrectly, I'll create a separate PR to fix all of them at once, so I'd leave this as is for now.

PR #54, opened 2019-05-21
Fix scope comparison

ajgarlag:
When a client with limited scopes requests a token with one of those scopes, current code fails because it tries to search an `Scope` model in an array of `Scope` models, instead of comparing its string value.


spideyfusion:
Thank you for your continued contributions @ajgarlag! :-)

PR #55, opened 2019-05-23
Command : Clean expired OAuth2 tokens 

anthony-launay:
New command to clean expired OAuth2 tokens


HypeMC:
@anthony-launay Thank you for the PR, but, this has already been resolved with https://github.com/trikoder/oauth2-bundle/pull/53

PR #56, opened 2019-05-29
Remove isRequired() from doctrine config

Nyholm:
This would allow me to write:

```yaml
trikoder_oauth2:
  persistence:
    doctrine: ~
```

If we do have `isRequired()` then there is no point in specifying a default value for `entity_manager`


X-Coder264 (approved):
Thanks for the PR @Nyholm.

Nyholm:
Thank you for merging

PR #62, opened 2019-06-04
Throw exceptions instead of creating responses

Nyholm:
We should throw exceptions instead of setting a response to the `GetResponseEvent`. This will give the developer a chance to format the response. Ie to support JSON reponses. 

I've also created a listener to keep the exisiting functionality.

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -92,9 +92,7 @@
             <argument type="service" id="Symfony\Component\Security\Core\Security" />
             <tag name="kernel.event_listener" event="trikoder.oauth2.authorization_request_resolve" method="onAuthorizationRequest" priority="1024" />
         </service>
-        <service id="trikoder.oauth2.event_listener.authorization.convert_to_response" class="Trikoder\Bundle\OAuth2Bundle\EventListener\ConvertExceptionToResponseListener">
-            <tag name="kernel.event_listener" event="kernel.exception" method="onKernelException" priority="10" />
-        </service>
+        <service id="trikoder.oauth2.event_listener.authorization.convert_to_response" class="Trikoder\Bundle\OAuth2Bundle\EventListener\ConvertExceptionToResponseListener" />
 
         <!-- Token controller -->
         <service id="trikoder.oauth2.controller.token_controller" class="Trikoder\Bundle\OAuth2Bundle\Controller\TokenController">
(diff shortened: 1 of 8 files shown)


Resources/config/services.xml
```diff
@@ -92,6 +92,9 @@
             <argument type="service" id="Symfony\Component\Security\Core\Security" />
             <tag name="kernel.event_listener" event="trikoder.oauth2.authorization_request_resolve" method="onAuthorizationRequest" priority="1024" />
         </service>
+        <service id="trikoder.oauth2.event_listener.authorization.convert_to_response" class="Trikoder\Bundle\OAuth2Bundle\EventListener\ConvertExceptionToResponseListener">
+            <tag name="kernel.event_listener" event="kernel.exception" method="onKernelException" priority="10" />
```
X-Coder264: @Nyholm Please make the priority configurable via the bundle extension. We've had already similar cases in our (legacy) apps where we had to override listener definitions from some bundles in our apps just to change the priority so I think it would be good to be flexible here.

Security/Exception/InsufficientScopesException.php
```diff
@@ -0,0 +1,22 @@
...
+
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Core\Exception\AuthenticationException;
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class InsufficientScopesException extends AuthenticationException
+{
+    public static function create(TokenInterface $token)
```
X-Coder264: Return type hint

Security/Exception/Oauth2AuthenticationFailedException.php
```diff
@@ -0,0 +1,18 @@
...
+namespace Trikoder\Bundle\OAuth2Bundle\Security\Exception;
+
+use Symfony\Component\Security\Core\Exception\AuthenticationException;
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class Oauth2AuthenticationFailedException extends AuthenticationException
+{
+    public static function create(string $message)
```
X-Coder264: Return type hint

Nyholm:
Thank you for the review. I've updated the PR

blry:
Hi @Nyholm 
I've noticed you throw exception from OAuth2Listener, however not from Token and Auth Controller. Why? Can't handle excepions from Controllers :(

PR #63, opened 2019-06-04
Support Defuse encryption

Nyholm:


DependencyInjection/Configuration.php
```diff
@@ -45,10 +45,15 @@ private function createAuthorizationServerNode(): NodeDefinition
                     ->defaultValue(null)
                 ->end()
                 ->scalarNode('encryption_key')
-                    ->info("The string used as an encryption key.\nHow to generate an encryption key: https://oauth2.thephpleague.com/installation/#string-password")
+                    ->info("The plain string or the ascii safe string used to create a \Defuse\Crypto\Key to be used as an encryption key.\nHow to generate an encryption key: https://oauth2.thephpleague.com/installation/#string-password")
```
X-Coder264: You can use here sprintf() and Key::class

X-Coder264:
@Nyholm Thank for the PR. Could you please add a bit of documentation for this so that people can see that this feature exists in the bundle? The reference configuration file in the `README.md` would also need to be updated to reflect the new state with the default values. Also it'd be good to add the `defuse/php-encryption` package in the composer.json `suggest` section.

Nyholm:
Updated.

Nyholm:
Thank you for merging.

PR #64, opened 2019-06-06
Various improvements

HypeMC:


Tests/Fixtures/FixtureFactory.php
```diff
@@ -56,6 +56,7 @@ final class FixtureFactory
     public const FIXTURE_SCOPE_SECOND = 'rock';
 
     public const FIXTURE_USER = 'user';
+    public const FIXTURE_PASSWORD = 'password';
```
X-Coder264: @HypeMC I don't see this being used anywhere?

HypeMC: @X-Coder264 It was add with https://github.com/trikoder/oauth2-bundle/pull/18/files#diff-a87d1acb0e93f0f5a6dd7abd960527e3 , the tests pass without it, but I figured having it doesn't hurt.

Tests/Unit/InMemoryRefreshTokenManagerTest.php
```diff
@@ -20,15 +20,17 @@ public function testClearExpired(): void
 
         timecop_freeze(new DateTime());
 
-        $testData = $this->buildClearExpiredTestData();
+        try {
+            $testData = $this->buildClearExpiredTestData();
```
X-Coder264: @HypeMC Does this line (and similar lines in other tests) need to be inside the `try` block? I don't see a case where building the data can throw an exception.

HypeMC: @X-Coder264 Probably not, in theory, `new DateTime` could throw an exception if an invalid value is passed, but this is more for consistency I guess, everything between free & return are in a try..finally block.

PR #65, opened 2019-06-11
Use DateTimeInterface instead of DateTime

carlos-ea:
This allows both `DateTime` and `DateTimeImmutable` as parameter types.


Nyholm:
I like this change. But we should update all the places we use DateTime.

carlos-ea:
Thank you @Nyholm . I have updated the PR.

spideyfusion:
Thank you for your contribution @carlos-ea!

PR #73, opened 2019-06-26
Clear expired tokens command fix

### LINKED ISSUE #72: Command clear-expired-tokens and foreign key constraint fails
This command makes no sense.
There will be exception "SQLSTATE[23000]: Integrity constraint violation:" when you try remove from access_token when there are related records in refresh_token table.
You can't remove access_token without removing refresh_token first, so you have to wait until refresh_token expires. 
So only command for removing from refresh_token works, but not from access_token.
This command should be changes so you can only run something like this:
- get access_token from refresh_token table where expired
- delete from refresh_token table where expired
- delete from access_token table where identity = access_token from refresh_token table

HypeMC:
Fixes #72 .

@spideyfusion Don't know why the `RefreshToken::$accessToken` wasn't nullable, but since `getAccessToken` is only used in some tests, making it nullable shouldn't be a problem. It makes sense since you can have a valid refresh token & an expired access token.

### DIFF
--- a/Tests/Acceptance/AbstractAcceptanceTest.php
+++ b/Tests/Acceptance/AbstractAcceptanceTest.php
@@ -28,5 +28,11 @@ abstract class AbstractAcceptanceTest extends WebTestCase
         $this->application = new Application($this->client->getKernel());
 
         TestHelper::initializeDoctrineSchema($this->application);
+
+        $connection = $this->client->getContainer()->get('database_connection');
+        if ('sqlite' === $connection->getDatabasePlatform()->getName()) {
+            // https://www.sqlite.org/foreignkeys.html
+            $connection->executeQuery('PRAGMA foreign_keys = ON');
+        }
     }
 }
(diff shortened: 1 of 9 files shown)


X-Coder264:
We should have a test that confirms that this works now and to prevent any future regressions.

HypeMC:
@X-Coder264 Done.

Tests/Acceptance/AbstractAcceptanceTest.php
```diff
@@ -28,5 +28,8 @@ protected function setUp(): void
         $this->application = new Application($this->client->getKernel());
 
         TestHelper::initializeDoctrineSchema($this->application);
+
+        // https://www.sqlite.org/foreignkeys.html
+        $this->client->getContainer()->get('database_connection')->executeQuery('PRAGMA foreign_keys = ON');
```
X-Coder264: @HypeMC As this is a sqlite specific query I think we should check first the platform of the connection via the `getDatabasePlatform` method and execute the query only if the platform is actually sqlite. What do you think?

HypeMC: @X-Coder264 Done

PR #75, opened 2019-06-26
Docker improvements

HypeMC:
~Still need to write some docs for this.~

Done.


dev/docker/Dockerfile
```diff
@@ -17,13 +17,18 @@ RUN apk add --update --no-cache --virtual .build-deps \
         xml \
         zip \
     && pecl install \
-        xdebug-2.7.0RC1 \
+        xdebug-2.7.2 \
```
X-Coder264: @HypeMC is there a reason why we have to have any version of xdebug hardcoded here? Why not just `pecl install xdebug` so that we always get the latest version?

The same question applies for the composer version too.

HypeMC: @X-Coder264 Done, I left composer as is like we talked

Issue #77, opened 2019-06-30
[Easy pick] Use FQCN for service ids


Nyholm:
All services should be using FQCN (fully qualified class names (namespace + class)) as their service ids. 

We could add aliases if we really wanted to keep our sf3.x standard. 

```yaml
    Trikoder\Bundle\OAuth2Bundle\Converter\UserConverter: '@trikoder.oauth2.converter.user_converter'
```

X-Coder264:
Fixed by https://github.com/trikoder/oauth2-bundle/pull/80

😃

PR #80, opened 2019-07-02
Replace service ids by their FQCN

rjwebdev:
Pull request for issue #77 

I replaced the service ID's by their FQCN and created aliasses for each service

### DIFF
--- a/DependencyInjection/TrikoderOAuth2Extension.php
+++ b/DependencyInjection/TrikoderOAuth2Extension.php
@@ -275,22 +275,22 @@ final class TrikoderOAuth2Extension extends Extension implements PrependExtensio
         );
 
         $container
-            ->getDefinition(AccessTokenManager::class)
+            ->getDefinition('trikoder.oauth2.manager.doctrine.access_token_manager')
             ->replaceArgument('$entityManager', $entityManager)
         ;
 
         $container
-            ->getDefinition(ClientManager::class)
+            ->getDefinition('trikoder.oauth2.manager.doctrine.client_manager')
             ->replaceArgument('$entityManager', $entityManager)
         ;
 
         $container
-            ->getDefinition(RefreshTokenManager::class)
+            ->getDefinition('trikoder.oauth2.manager.doctrine.refresh_token_manager')
             ->replaceArgument('$entityManager', $entityManager)
         ;
 
         $container
-            ->getDefinition(AuthorizationCodeManager::class)
+            ->getDefinition('trikoder.oauth2.manager.doctrine.authorization_code_manager')
             ->replaceArgument('$entityManager', $entityManager)
         ;
 
--- a/DependencyInjection/TrikoderOAuth2Extension.php
+++ b/DependencyInjection/TrikoderOAuth2Extension.php
@@ -275,22 +275,22 @@ final class TrikoderOAuth2Extension extends Extension implements PrependExtensio
         );
 
         $container
-            ->getDefinition('trikoder.oauth2.manager.doctrine.access_token_manager')
+            ->getDefinition(AccessTokenManager::class)
             ->replaceArgument('$entityManager', $entityManager)
         ;
 
         $container
-            ->getDefinition('trikoder.oauth2.manager.doctrine.client_manager')
+            ->getDefinition(ClientManager::class)
             ->replaceArgument('$entityManager', $entityManager)
         ;
 
         $container
-            ->getDefinition('trikoder.oauth2.manager.doctrine.refresh_token_manager')
+            ->getDefinition(RefreshTokenManager::class)
             ->replaceArgument('$entityManager', $entityManager)
         ;
 
         $container
-            ->getDefinition('trikoder.oauth2.manager.doctrine.authorization_code_manager')
+            ->getDefinition(AuthorizationCodeManager::class)
             ->replaceArgument('$entityManager', $entityManager)
         ;
 
--- a/DependencyInjection/TrikoderOAuth2Extension.php
+++ b/DependencyInjection/TrikoderOAuth2Extension.php
@@ -7,7 +7,14 @@ namespace Trikoder\Bundle\OAuth2Bundle\DependencyInjection;
 use DateInterval;
 use Defuse\Crypto\Key;
 use Doctrine\Bundle\DoctrineBundle\DoctrineBundle;
+use League\OAuth2\Server\AuthorizationServer;
 use League\OAuth2\Server\CryptKey;
+use League\OAuth2\Server\Grant\AuthCodeGrant;
+use League\OAuth2\Server\Grant\ClientCredentialsGrant;
+use League\OAuth2\Server\Grant\ImplicitGrant;
+use League\OAuth2\Server\Grant\PasswordGrant;
+use League\OAuth2\Server\Grant\RefreshTokenGrant;
+use League\OAuth2\Server\ResourceServer;
 use LogicException;
 use Psr\Http\Message\ResponseFactoryInterface;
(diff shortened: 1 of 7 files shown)


X-Coder264:
@rjwebdev Thank you for the PR. Please also add deprecation messages on these aliases stating that we will remove them in the next major version (3.0).

DependencyInjection/TrikoderOAuth2Extension.php
```diff
@@ -48,7 +48,7 @@ public function load(array $configs, ContainerBuilder $container)
         $this->configureResourceServer($container, $config['resource_server']);
         $this->configureScopes($container, $config['scopes']);
 
-        $container->getDefinition('trikoder.oauth2.event_listener.authorization.convert_to_response')
+        $container->getDefinition('Trikoder\Bundle\OAuth2Bundle\EventListener\ConvertExceptionToResponseListener')
```
X-Coder264: @rjwebdev You can use now in the entire extension the `::class` notation instead of a string like this.

rjwebdev: Good point! Wanted to be to fast, will replace it

X-Coder264:
Sorry for bothering you to add the deprecation messages to the aliases. I forgot that this is a new feature that was actually added only recently in Symfony 4.3 and we support 3.4+ so please remove them.

rjwebdev:
@X-Coder264 I removed the alias deprecation warnings and changed the notations for the class names in the extension.

Pull request good this way?

X-Coder264:
@rjwebdev From what I see from the diff you didn't replace to `::class` in all places.

DependencyInjection/TrikoderOAuth2Extension.php
HypeMC: @rjwebdev You seemed to have missed this one.

DependencyInjection/TrikoderOAuth2Extension.php
```diff
@@ -167,35 +174,36 @@ private function configureAuthorizationServer(ContainerBuilder $container, array
...
+                new Reference(AuthCodeGrant::class),
                 new Definition(DateInterval::class, [$config['access_token_ttl']]),
             ]);
         }
 
         if ($config['enable_implicit_grant']) {
             $authorizationServer->addMethodCall('enableGrantType', [
-                new Reference('league.oauth2.server.grant.implicit_grant'),
+                new Reference('League\OAuth2\Server\Grant\ImplicitGrant'),
+                new Reference(ImplicitGrant::class),
```
X-Coder264: @rjwebdev Here you have now the ImplictGrant two times, on this line and on the one above.

DependencyInjection/TrikoderOAuth2Extension.php
```diff
@@ -167,35 +174,36 @@ private function configureAuthorizationServer(ContainerBuilder $container, array
...
-                new Reference('league.oauth2.server.grant.auth_code_grant'),
+                new Reference(AuthCodeGrant::class),
                 new Definition(DateInterval::class, [$config['access_token_ttl']]),
             ]);
         }
 
         if ($config['enable_implicit_grant']) {
             $authorizationServer->addMethodCall('enableGrantType', [
-                new Reference('league.oauth2.server.grant.implicit_grant'),
+                new Reference('League\OAuth2\Server\Grant\ImplicitGrant'),
```
HypeMC: @rjwebdev You can use `::class` here as well.

Resources/config/services.xml
```diff
@@ -4,134 +4,183 @@
     <services>
         <!-- The league interfaces -->
         <service id="League\OAuth2\Server\AuthorizationServer" alias="league.oauth2.server.authorization_server" />
+        <service id="league.oauth2.server.authorization_server" alias="League\OAuth2\Server\AuthorizationServer" />
```
HypeMC: @rjwebdev This seems to be a duplicate, you already have this on line 78.

Resources/config/services.xml
```diff
@@ -4,134 +4,183 @@
...
+
         <service id="League\OAuth2\Server\ResourceServer" alias="league.oauth2.server.resource_server" />
+        <service id="league.oauth2.server.resource_server" alias="League\OAuth2\Server\ResourceServer" />
```
HypeMC: @rjwebdev Same here, line 85.

rjwebdev: Should be fixed

Resources/config/services.xml
rjwebdev: Should be removed.

DependencyInjection/TrikoderOAuth2Extension.php
rjwebdev: Sorry, again too fast. Removed first line

HypeMC:
@rjwebdev Hi, there are still several issues with this PR:
1) the CI fails cause the code is not linted properly
2) there are still some definitions in `Resources/config/storage/*.xml` that need to be switched to their FQCN
3) there are a couple of new deprecations:
```
  7x: The "trikoder.oauth2.controller.authorization_controller" service is private, getting it from the container is deprecated since Symfony 3.2 and will fail in 4.0. You should either make the service public, or stop using the container directly and use dependency injection instead.
  6x: The "trikoder.oauth2.controller.token_controller" service is private, checking for its existence is deprecated since Symfony 3.2 and will fail in 4.0.
```

You'll need to set the `trikoder.oauth2.controller.*` aliases to public

Resources/config/services.xml
```diff
@@ -2,136 +2,179 @@
            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
-        <!-- The league interfaces -->
-        <service id="League\OAuth2\Server\AuthorizationServer" alias="league.oauth2.server.authorization_server" />
-        <service id="League\OAuth2\Server\ResourceServer" alias="league.oauth2.server.resource_server" />
```
HypeMC: @rjwebdev Nitpick, no need for this extra line

ajgarlag:
>     3. there are a couple of new deprecations:
> 
> 
> ```
>   7x: The "trikoder.oauth2.controller.authorization_controller" service is private, getting it from the container is deprecated since Symfony 3.2 and will fail in 4.0. You should either make the service public, or stop using the container directly and use dependency injection instead.
>   6x: The "trikoder.oauth2.controller.token_controller" service is private, checking for its existence is deprecated since Symfony 3.2 and will fail in 4.0.
> ```
> 
> You'll need to set the `trikoder.oauth2.controller.*` aliases to public

Not sure, but I think this error can be solved using the FQCN in routes.xml

HypeMC:
@rjwebdev Even better, if it can be solved like that then great.

rjwebdev:
@X-Coder264 It looks like everything is fixed at the moment

Nyholm (approved):
This was way more changes than I originally thought. 

Great job!

X-Coder264:
@rjwebdev 

In OAuth2Factory `'trikoder.oauth2.security.entry_point.oauth2_entry_point'` can also be replaced with the FQCN.

Some services are still wired to get the `trikoder.oauth2.converter.scope_converter` alias instead of the FQCN ID (the same is the case for one service being wired with `trikoder.oauth2.event.authorization_request_resolve_event_factory`)

The rest looks fine now.

X-Coder264:
@rjwebdev Thanks for the PR.

Nyholm:
Wohoo. Awesome work!

Issue #76, opened 2019-06-30
[Easy pick] All final classes must have an interface


Nyholm:
We need this to be able to mock and tests our code.

X-Coder264:
I don't think we should blindly add interfaces to all final classes in the bundle just for some "test" purposes. For example I don't see any value in adding an interface for the `UserConverter` class. I don't see the possibility of there being another implementation that justifies an interface (or it's very unlikely at least) nor does this implementation do any I/O or database calls etc which means doing `new UserConverter()` in your tests is just fine if you ask me.

Nyholm:
No, not "blindly". Final classes that are not "internal" should have an interface. 

Speaking of your example of `UserConverter`: 
I like the `OauthAuthorizationController` but I need do to some modifications to it. So I copied that class to my namespace and did my modifications. (Everything is perfect.) However, when I write unit tests from my new controller class I cannot mock its dependencies. That is reason 1 why we need an interface. 

Reason 2 is that `OauthAuthorizationController` has dependency on a concrete class. That violates the dependency inversion principle. 

---------

You are probably correct when you say that there will most likely not be an other implementation of the `UserConverter`. However, adding an interface to be more correct will guarantee future proof code.

alexspark21:
Yep, I think we have to set UserConverterInterface,  because I have to have possibility override UserConverter and implement my logic

janvernieuwe:
I just ran into a situation where a `ScopeInterface` would have been nice.
When retrieving scopes from a custom system you could just implement it and return them.

X-Coder264:
I think everything here that was mentioned is solved. Feel free to open a new issue if there are still things that need to be improved.

PR #82, opened 2019-07-06
Adding AuthorizationRequestResolveEventInterface

Nyholm:
Related to #76 

Changes: 
- Adding an interface to our 3 events.
- Remove `return $this;`
- ~~Use `EventDispatcherInterface` from symfony contracts.~~
- Using the new syntax for `EventDispatcherInterface->dispatch()` which have the parameter swapped places.

### DIFF
--- a/Event/AuthorizationRequestResolveEvent.php
+++ b/Event/AuthorizationRequestResolveEvent.php
@@ -13,13 +13,9 @@ use Symfony\Component\Security\Core\User\UserInterface;
 use Trikoder\Bundle\OAuth2Bundle\Converter\ScopeConverter;
 use Trikoder\Bundle\OAuth2Bundle\Manager\ClientManagerInterface;
 use Trikoder\Bundle\OAuth2Bundle\Model\Client;
-use Trikoder\Bundle\OAuth2Bundle\Model\Scope;
 
-final class AuthorizationRequestResolveEvent extends Event
+final class AuthorizationRequestResolveEvent extends Event implements AuthorizationRequestResolveEventInterface
 {
-    public const AUTHORIZATION_APPROVED = true;
-    public const AUTHORIZATION_DENIED = false;
-
     /**
      * @var AuthorizationRequest
      */
@@ -38,7 +34,7 @@ final class AuthorizationRequestResolveEvent extends Event
     /**
      * @var bool
      */
-    private $authorizationResolution = self::AUTHORIZATION_DENIED;
+    private $authorizationResolution = AuthorizationRequestResolveEventInterface::AUTHORIZATION_DENIED;
 
     /**
      * @var ResponseInterface|null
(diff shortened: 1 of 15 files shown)


Nyholm:
Hm, Switching to EventDispatcherInterface from symfony contracts will break for 3.4 and 4.2 because FramworkBundle does not support it. 

I will revert this and add that change in another PR.

Event/UserResolveEventInterface.php
```diff
@@ -0,0 +1,30 @@
...
+use Trikoder\Bundle\OAuth2Bundle\Model\Client;
+use Trikoder\Bundle\OAuth2Bundle\Model\Grant;
+
+/**
+ * Thi event occurs when the client requests a "password"
+ * grant type from the authorization server.
+ *
+ * You should set a valid user here if applicable.
+ */
+interface UserResolveEventInterface
```
HypeMC: @Nyholm Any reason you didn't update `Tests/Integration/AuthorizationServerTest.php` & `Tests/Acceptance/TokenEndpointTest.php` to use the interface?

Also, `docs/password-grant-handling.md` should be updated as well.

Nyholm:
Rebased and ready for review

Event/AuthorizationRequestResolveEvent.php
```diff
@@ -62,13 +58,11 @@ public function getAuthorizationResolution(): bool
...
     }
 
-    public function resolveAuthorization(bool $authorizationResolution): self
+    public function resolveAuthorization(bool $authorizationResolution): void
     {
         $this->authorizationResolution = $authorizationResolution;
         $this->response = null;
         $this->stopPropagation();
-
-        return $this;
```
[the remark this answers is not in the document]
Nyholm: Fluent interfaces should generally not be used. Marco Pivetta writes it down real nice at: https://ocramius.github.io/blog/fluent-interfaces-are-evil/

Of course there are exceptions. Like if you doing a "builder".

Also, we never used this ourselves.

Event/UserResolveEventInterface.php
Nyholm: Nope, I just missed them.

Nyholm:
Friendly ping

spideyfusion:
@Nyholm Sorry for the late reply. We've all been pretty much in vacation mode this past month. We'll take a look at pending PRs sometime this weekend.

Nyholm:
That is great. Thank you.

spideyfusion:
@Nyholm I agree that we should interface our service classes (e.g. converters, factories), but can you explain me the benefits of interfacing event objects?

I noticed that currently our `AuthorizationRequestResolveEvent` event is rather complex to instantiate (it needs a `ScopeConverter` instance and a `ClientManagerInterface` implementation). We could simplify it by completely delegating the work to `AuthorizationRequestResolveEventFactory` so that the event object already consumes prepared data from the factory. What do you think?

Generally, event (value) objects shouldn't have a need to be mocked in tests as they should contain no complex logic and be easy to instantiate.

Nyholm:
I use this event all over my source code. I replaced some of the bundle's listeners with my own because I needed to make a few changes. 
I would really like to be able to test my code in a simple way. With an interface on `AuthorizationRequestResolveEvent` I could mock that interface and not bother about the complex way to construct this object. 

About moving some of the code into a factory: 
I like that idea. I think this should be merge and we can easily do the refactor in an other PR.

X-Coder264:
@Nyholm Thanks for the PR, but it was superseded by https://github.com/trikoder/oauth2-bundle/pull/125 which made creating an `AuthorizationRequestResolveEvent` object easier (thus making the need for an interface obsolete).

PR #83, opened 2019-07-06
Use Symfony Flex so better specify symfony version

Nyholm:
This will produce the same output. The only difference is: 
- .This one is quicker
- This will not install symfony/symfony. Ie, we will see if we are missing dependencies.


X-Coder264:
Thanks for the PR.

PR #85, opened 2019-07-06
Make sure tests on different sf versions are wokring

Nyholm:
I made a misstake in #83.
Using `--no-scripts` will disable flex from running. In current master we are always using latest sf version. 
I wanted to run `--no-scripts` to avoid recipes being installed. That would add a bunch of new files that does not comply with our CS guidelines, ie build will always fail.

This PR first installs flex globally. Then install all packages to the correct (lower) version. That means no recipes are installed. 

~~I also use environment variable instad.~~

### DIFF
--- a/dev/docker/Dockerfile
+++ b/dev/docker/Dockerfile
@@ -47,7 +47,7 @@ RUN curl --show-error https://getcomposer.org/installer | php -- \
 RUN useradd -ms /bin/sh app
 
 # Enable parallel package installation for Composer
-RUN su-exec app composer global require hirak/prestissimo
+RUN su-exec app composer global require symfony/flex
 
 COPY entrypoint.sh /usr/local/bin/docker-entrypoint
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,10 @@ install:
     - dev/bin/docker-compose build --build-arg PHP_VERSION=${PHP_VERSION} php
 
 before_script:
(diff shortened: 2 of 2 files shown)


X-Coder264:
It's still always pulling in 4.3.2 and ignoring the version that's being set. And why does composer update have to be called twice? All the second call does is output the "Nothing to install or update" message.

Nyholm:
Yeah, Im still debugging. I think I found a solution now. I install flex globally. 

Waiting on travis with suspense =)

X-Coder264:
It still doesn't work :stuck_out_tongue:

Nyholm:
Hm... It works locally.. Strange. I'll keep trying for an hour or so.

Nyholm:
Okey, I think I know that the issue was. Since composer is in the docker image it never saw my environment variable. =)

dev/docker/Dockerfile
```diff
@@ -47,7 +47,7 @@ RUN curl --show-error https://getcomposer.org/installer | php -- \
 RUN useradd -ms /bin/sh app
 
 # Enable parallel package installation for Composer
-RUN su-exec app composer global require hirak/prestissimo
```
Nyholm: Flex is also installing packages in parallel and it supports specifying symfony versions

Nyholm:
Wohoo. It is working now =)

Nyholm:
Thank you for merging

PR #87, opened 2019-07-08
Remove deprecation notice from sf 4.3

Nyholm:
I added a BC layer so we are still compatible with sf 3.4


spideyfusion:
@Nyholm Thanks for the PR! 🎊

Nyholm:
Thank you for merging

PR #90, opened 2019-08-07
Add aliases for League interfaces

X-Coder264:
This PR adds aliases for the League interfaces so that these interfaces can be autowired. These aliases are also used now in all our other services (instead of them being hardcoded to our implementation) so that the user can easily swap the interface implementation in all services by aliasing them in their app.


spideyfusion:
@X-Coder264 Thanks for the improvement! 🎉

Issue #16, opened 2019-01-08
Add config option that allows us to override role name prefix


miroslav-marinkovic:
Currently, role names are prefixed with `ROLE_OAUTH2_`. A config option to override this would make bundle a lot more flexible.

rjwebdev:
Should this role name be configured in the firewall or in the bundle configuration?

spideyfusion:
@rjwebdev Definitely in the bundle configuration.

HypeMC:
Closed with #94.

PR #94, opened 2019-08-09
Add custom_role config option

rjwebdev:
Change for issue #16 

With this change, it is possible to define a custom role prefix that will replace the standard 'ROLE_OAUTH2_' prefix.

Any thoughts on this?

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -28,14 +28,14 @@
             <argument type="service" id="Trikoder\Bundle\OAuth2Bundle\Manager\ScopeManagerInterface" />
             <argument type="service" id="Trikoder\Bundle\OAuth2Bundle\Manager\ClientManagerInterface" />
             <argument type="service" id="Trikoder\Bundle\OAuth2Bundle\Converter\ScopeConverter" />
-            <argument type="service" id="Symfony\Component\EventDispatcher\EventDispatcherInterface" />
+            <argument type="service" id="Symfony\Contracts\EventDispatcher\EventDispatcherInterface" />
         </service>
         <service id="trikoder.oauth2.league.repository.scope_repository" alias="Trikoder\Bundle\OAuth2Bundle\League\Repository\ScopeRepository" />
         <service id="League\OAuth2\Server\Repositories\ScopeRepositoryInterface" alias="Trikoder\Bundle\OAuth2Bundle\League\Repository\ScopeRepository" />
 
         <service id="Trikoder\Bundle\OAuth2Bundle\League\Repository\UserRepository">
             <argument type="service" id="Trikoder\Bundle\OAuth2Bundle\Manager\ClientManagerInterface" />
-            <argument type="service" id="Symfony\Component\EventDispatcher\EventDispatcherInterface" />
+            <argument type="service" id="Symfony\Contracts\EventDispatcher\EventDispatcherInterface" />
             <argument type="service" id="Trikoder\Bundle\OAuth2Bundle\Converter\UserConverter" />
         </service>
         <service id="trikoder.oauth2.league.repository.user_repository" alias="Trikoder\Bundle\OAuth2Bundle\League\Repository\UserRepository" />
@@ -113,7 +113,7 @@
         <!-- Authorization controller -->
         <service id="Trikoder\Bundle\OAuth2Bundle\Controller\AuthorizationController">
             <argument type="service" id="League\OAuth2\Server\AuthorizationServer" />
-            <argument type="service" id="Symfony\Component\EventDispatcher\EventDispatcherInterface" />
+            <argument type="service" id="Symfony\Contracts\EventDispatcher\EventDispatcherInterface" />
             <argument type="service" id="Trikoder\Bundle\OAuth2Bundle\Event\AuthorizationRequestResolveEventFactory" />
             <argument type="service" id="Trikoder\Bundle\OAuth2Bundle\Converter\UserConverter" />
             <tag name="controller.service_arguments" />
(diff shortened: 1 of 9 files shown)


DependencyInjection/Configuration.php
```diff
@@ -30,6 +30,9 @@ public function getConfigTreeBuilder()
                     ->info('The priority of the event listener that converts an Exception to a Response')
                     ->defaultValue(10)
                 ->end()
+                ->scalarNode('custom_role')
+                    ->info('Custom role that replaces \'ROLE_OAUTH2_\' role prefix')
```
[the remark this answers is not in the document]
rjwebdev: Created a factory that is injected into the services where the `OAuth2Token` constructor was called so that the token will be created in this factory.

Also rebased and retargeted

DependencyInjection/Configuration.php
```diff
@@ -30,6 +30,10 @@ public function getConfigTreeBuilder()
                     ->info('The priority of the event listener that converts an Exception to a Response')
                     ->defaultValue(10)
                 ->end()
+                ->scalarNode('custom_role')
```
spideyfusion: @rjwebdev Keep the naming consistent throughout the code. I personally find `role_prefix` to be more descriptive. :)

Resources/config/services.xml
```diff
@@ -53,6 +53,7 @@
         <service id="Trikoder\Bundle\OAuth2Bundle\Security\Authentication\Provider\OAuth2Provider">
             <argument key="$userProvider" />
             <argument key="$resourceServer" type="service" id="League\OAuth2\Server\ResourceServer" />
+            <argument key="$oauth2TokenFactory" type="service" id="Trikoder\Bundle\OAuth2Bundle\Security\Authentication\Token\OAuth2TokenFactory"
```
spideyfusion: @rjwebdev You forgot to self-close the tag, whoops!

Security/Authentication/Token/OAuth2Token.php
```diff
@@ -38,8 +39,8 @@ private function buildRolesFromScopes(): array
     {
         $roles = [];
 
-        foreach ($this->getAttribute('server_request')->getAttribute('oauth_scopes', []) as $scope) {
```
spideyfusion: @rjwebdev This seems like an accidental change. The attribute on the server request is really named `oauth_scopes`.

Security/Authentication/Token/OAuth2Token.php
```diff
@@ -38,8 +39,8 @@ private function buildRolesFromScopes(): array
...
-            $roles[] = sprintf('ROLE_OAUTH2_%s', trim(strtoupper($scope)));
+        foreach ($this->getAttribute('server_request')->getAttribute('oauth2_scopes', []) as $scope) {
+            $roles[] = sprintf($this->getAttribute('role_prefix') . '%s', trim(strtoupper($scope)));
```
spideyfusion:
```suggestion
            $roles[] = $this->getAttribute('role_prefix') . trim(strtoupper($scope)));
```

Security/Authentication/Token/OAuth2TokenFactory.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Trikoder\Bundle\OAuth2Bundle\Security\Authentication\Token;
+
+use Psr\Http\Message\ServerRequestInterface;
+use Symfony\Component\Security\Core\User\UserInterface;
+
+class OAuth2TokenFactory
```
spideyfusion: @rjwebdev Can we make the class `final`?

Security/Authentication/Token/OAuth2TokenFactory.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    /**
+     * @var string
+     */
+    private $rolePrefix;
+
+    /**
+     * OAuth2TokenFactory constructor.
```
spideyfusion: @rjwebdev Redundant comment. :)

Security/Authentication/Token/OAuth2TokenFactory.php
```diff
@@ -0,0 +1,29 @@
...
+     */
+    public function __construct(string $rolePrefix)
```
spideyfusion: @rjwebdev I noticed that you made the role prefix on the `OAuth2Token` value object nullable, but here it is not. We need to be consistent.

A follow-up question: Do we want to allow the users to omit the prefix entirely?

Depending on which route we decide to take we'll have to add either `cannotBeEmpty` or `treatNullLike` options to the configuration node.

DependencyInjection/Configuration.php
rjwebdev: Changed

Resources/config/services.xml
rjwebdev: Closed the tag

Security/Authentication/Token/OAuth2Token.php
rjwebdev: Sorry, it was indeed by accident. Changed it back

Security/Authentication/Token/OAuth2Token.php
rjwebdev: Changed it to your suggestion

Security/Authentication/Token/OAuth2TokenFactory.php
rjwebdev: I made the class final. In my opinion, I don't think classes used in a bundle should be made final since there's always a chance someone who's using the  bundle, wants to override pieces of it

Security/Authentication/Token/OAuth2TokenFactory.php
rjwebdev: Removed it

Security/Authentication/Token/OAuth2TokenFactory.php
rjwebdev: Didn't remove the optional parameter in the `OAuth2Token` constructor

Not sure what you mean with your follow-up question

spideyfusion:
@rjwebdev Sorry for messing with the forked branch. Your feature will be integrated into the bundle via https://github.com/trikoder/oauth2-bundle/pull/122.

PR #122, opened 2019-11-28
Add custom_role config option

Allypost:
Fixes #94

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -30,6 +30,9 @@ final class Configuration implements ConfigurationInterface
                     ->info('The priority of the event listener that converts an Exception to a Response')
                     ->defaultValue(10)
                 ->end()
+                ->scalarNode('custom_role')
+                    ->info('Custom role that replaces \'ROLE_OAUTH2_\' role prefix')
+                ->end()
             ->end();
 
         return $treeBuilder;
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -34,6 +34,7 @@ final class Configuration implements ConfigurationInterface
                 ->scalarNode('role_prefix')
                     ->info('Custom role that replaces \'ROLE_OAUTH2_\' role prefix')
                     ->defaultValue('ROLE_OAUTH2_')
+                    ->cannotBeEmpty()
                 ->end()
             ->end();
 
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -14,10 +14,9 @@ final class Configuration implements ConfigurationInterface
     /**
      * {@inheritdoc}
      */
-    public function getConfigTreeBuilder()
+    public function getConfigTreeBuilder(): TreeBuilder
     {
         $treeBuilder = $this->getWrappedTreeBuilder('trikoder_oauth2');
-        /** @var NodeDefinition $rootNode */
         $rootNode = $treeBuilder->getRootNode();
 
         $rootNode->append($this->createAuthorizationServerNode());
@@ -173,7 +172,7 @@ final class Configuration implements ConfigurationInterface
         return $node;
     }
 
-    private function getWrappedTreeBuilder(string $name): object
+    private function getWrappedTreeBuilder(string $name): TreeBuilder
     {
         return new class($name) extends TreeBuilder {
             public function __construct(string $name)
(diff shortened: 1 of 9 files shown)


Security/Authentication/Token/OAuth2Token.php
```diff
@@ -39,7 +40,7 @@ private function buildRolesFromScopes(): array
         $roles = [];
 
         foreach ($this->getAttribute('server_request')->getAttribute('oauth_scopes', []) as $scope) {
-            $roles[] = sprintf('ROLE_OAUTH2_%s', trim(strtoupper($scope)));
+            $roles[] = $this->getAttribute('role_prefix') . trim(strtoupper($scope));
```
spideyfusion: @Allypost You should ensure that the entire role is now uppercase since now we're giving the developer an option to customize the prefix.

Allypost: @spideyfusion Should be fixed now. The role should now also be trimmed as a whole instead of just trimming the `scope` and leaving the `role_prefix` filthy with whitespace.

Resources/config/services.xml
```diff
@@ -183,5 +185,10 @@
             <argument type="service" id="Trikoder\Bundle\OAuth2Bundle\Manager\ClientManagerInterface" />
         </service>
         <service id="trikoder.oauth2.event.authorization_request_resolve_event_factory" alias="Trikoder\Bundle\OAuth2Bundle\Event\AuthorizationRequestResolveEventFactory" />
+
+        <service id="Trikoder\Bundle\OAuth2Bundle\Security\Authentication\Token\OAuth2TokenFactory">
+            <argument type="string" />
+        </service>
+        <service id="trikoder.oauth2.security.authentication.token.oauth2_token_factory" alias="Trikoder\Bundle\OAuth2Bundle\Security\Authentication\Token\OAuth2TokenFactory" />
```
HypeMC: @Allypost @spideyfusion Wouldn't add legacy aliases to new classes.

README.md
```diff
@@ -96,6 +96,10 @@ This package is currently in the active development.
         # Scopes that you wish to utilize in your application.
         # This should be a simple array of strings.
         scopes:               []
+   
+        # The default role prefix for OAUTH roles that
+        # are generated from the set scopes
+        role_prefix:          'ROLE_OAUTH2_'
```
HypeMC: @Allypost @spideyfusion Don't think this was generated with `bin/console config:dump-reference` since the comment is different from the info.

Security/Authentication/Token/OAuth2Token.php
```diff
@@ -36,10 +37,11 @@ public function getCredentials()
 
     private function buildRolesFromScopes(): array
     {
+        $prefix = $this->getAttribute('role_prefix');
         $roles = [];
 
         foreach ($this->getAttribute('server_request')->getAttribute('oauth_scopes', []) as $scope) {
-            $roles[] = sprintf('ROLE_OAUTH2_%s', trim(strtoupper($scope)));
+            $roles[] = trim(strtoupper($prefix . $scope));
```
HypeMC: @Allypost @spideyfusion Should we ensure that `$prefix` ends with `_`?

HypeMC: @Allypost @spideyfusion Also, we should probably trim just the `$prefix` variable.

Allypost: @HypeMC I think the prefix should be freely user-definable (eg. if someone wants to separate their roles with a `:` they should be able to)

Allypost: @HypeMC I think the trimming bit is a bit of a micro optimisation (and IMHO it looks slightly cleaner this way)

Resources/config/services.xml
Allypost: @HypeMC Fixed

README.md
Allypost: @HypeMC Fixed

DependencyInjection/Configuration.php
```diff
@@ -17,6 +17,7 @@ final class Configuration implements ConfigurationInterface
     public function getConfigTreeBuilder()
     {
         $treeBuilder = $this->getWrappedTreeBuilder('trikoder_oauth2');
+        /** @var NodeDefinition $rootNode */
```
HypeMC: @Allypost Just noticed this, I think a better solution would be to fix the return type of `getWrappedTreeBuilder`, should be `TreeBuilder`.

Allypost: @HypeMC Fixed.

PR #125, opened 2019-11-29
Simplify AuthorizationRequestResolveEvent creation

spideyfusion:


Event/AuthorizationRequestResolveEvent.php
```diff
@@ -50,11 +47,11 @@ final class AuthorizationRequestResolveEvent extends Event
      */
     private $user;
 
-    public function __construct(AuthorizationRequest $authorizationRequest, ScopeConverterInterface $scopeConverter, ClientManagerInterface $clientManager)
+    public function __construct(AuthorizationRequest $authorizationRequest, array $scopes, Client $client)
```
HypeMC: @spideyfusion DO we want/need something like
```php
public function __construct(AuthorizationRequest $authorizationRequest, Client $client, Scope ...$scopes)
```

spideyfusion: @HypeMC I would say no because only the last parameter in a method signature can be declared as variadic. That's not really flexible and makes the contract fairly loose.

Although, I did add a phpDoc block to make the requirements more obvious. :-)

PR #136, opened 2019-12-05
[Feature] Ask user for confirmation on delete client

Allypost:


Manager/AccessTokenManagerInterface.php
```diff
@@ -6,11 +6,12 @@
...
 {
     public function find(string $identifier): ?AccessToken;
 
-    public function save(AccessToken $accessToken): void;
-
-    public function clearExpired(): int;
+    /**
+     * @param AccessToken $accessToken
+     */
+    public function save(/* AccessToken */ $accessToken): void;
```
X-Coder264: @Allypost This is a breaking change so this can't go to the v2 branch.

Issue #138, opened 2019-12-06
Support for PKCE workflow ?


florianajir:
As mentionned in this issue comment https://github.com/trikoder/oauth2-bundle/issues/70#issuecomment-504399190 the best practice for mobile app is authenticate without client_secret.
The `league/oauth2-server` v7.4 used by this bundle design does not allow PKCE extension by default due to `getClientEntity::$mustValidateSecret` set by `AbstractGrant` instance https://github.com/thephpleague/oauth2-server/blob/7.4.0/src/Grant/AbstractGrant.php#L194

resource: https://www.oauth.com/oauth2-servers/pkce/

florianajir:
According to `league/oauth2-server` [documentation ](https://oauth2.thephpleague.com/upgrade-guide/#client-entity-interface) this feature as been enabled by default in v8 to recognize in the request the need to switch the `$enableCodeExchangeProof` flag. 
So maybe only upgrading the dependency could help with this issue.

florianajir:
PKCE support available from master branch

PR #99, opened 2019-08-17
Support "league/oauth2-server" v8.x

spideyfusion:

PR #139, opened 2019-12-06
upgrade `league/oauth2-server` to `^8.0`

florianajir:
Upgrade `league/oauth2-server` to `^8.0` to allow PKCE extension.
Changelog: https://github.com/thephpleague/oauth2-server/releases/tag/8.0.0

fix https://github.com/trikoder/oauth2-bundle/issues/138

### DIFF
--- a/Command/CreateClientCommand.php
+++ b/Command/CreateClientCommand.php
@@ -57,6 +57,13 @@ final class CreateClientCommand extends Command
                 'Sets allowed scope for client. Use this option multiple times to set multiple scopes.',
                 []
             )
+            ->addOption(
+                'confidential',
+                null,
+                InputOption::VALUE_NONE,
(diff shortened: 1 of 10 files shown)


X-Coder264:
Thank you for the PR, but we have already done this (https://github.com/trikoder/oauth2-bundle/pull/99) (only) on master (which is gonna be released as 3.0 soon).

The reason why it's only on master is because as you can see from your PR there are quite some changes that had to be done (and some of them like interface method signature changes are breaking changes) so it couldn't be released in a minor version.

Also we wanted to be more careful with bumping a major version of the package that this bundle is a wrapper for so that we don't unintentionally break any user app on 2.x.

If there are some additional changes that must be made in order to properly support PKCE please create a new PR targeting the `master` branch.

PR #140, opened 2019-12-07
add Client::confidential property

florianajir:
add `Client::confidential` property to fit with PKCE client validation rule:
> If the client is confidential (i.e. is capable of securely storing a secret) then the secret must be validated.

https://oauth2.thephpleague.com/client-repository-interface/#validateclient--bool

### DIFF
--- a/Command/UpdateClientCommand.php
+++ b/Command/UpdateClientCommand.php
@@ -67,7 +67,7 @@ final class UpdateClientCommand extends Command
                 'confidential',
                 null,
                 InputOption::VALUE_OPTIONAL,
-                'Set client confidentiality. Use this option with value 0 to set client public.',
+                'Set client confidential. If the client is confidential then the secret must be validated (PKCE client validation).',
                 true
             )
             ->addArgument(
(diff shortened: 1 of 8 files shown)


Command/UpdateClientCommand.php
```diff
@@ -92,6 +99,10 @@ private function updateClientFromInput(Client $client, InputInterface $input): C
     {
         $client->setActive(!$input->getOption('deactivated'));
 
+        if ($input->hasOption('confidential')) {
```
HypeMC: @florianajir Keep in mind that `$input->hasOption` check whether the option has been added to the definition (with `addOption`) and **not** if it has been passed during execution (`--confidential`), so i don't think you're getting the expected behavior here.

Resources/config/doctrine/model/Client.orm.xml
```diff
@@ -9,5 +9,6 @@
         <field name="grants" type="oauth2_grant" nullable="true" />
         <field name="scopes" type="oauth2_scope" nullable="true" />
         <field name="active" type="boolean" />
+        <field name="confidential" type="boolean" />
```
X-Coder264: @florianajir When Laravel Passport implemented this feature they just made the secret column nullable and then `isConfidential === null !== $this->secret`

https://github.com/laravel/passport/pull/1065

I'm wondering what are the pros and cons of both approaches?

florianajir: @X-Coder264 Good point, the only pros I see is the ability to switch from public to confidential without regenerating the secret manually but on the cons the need of having a new data attribute.
It's quite the same IMO, if we go for the nullable secret then the actual code needs more changes.

X-Coder264: @florianajir I don't see the point of having a non confidential client with a secret as the Oauth specification says that non confidential / public clients are not capable of securely storing a secret. So having this extra column seems redundant.

X-Coder264:
@florianajir Thanks for the PR, but it was superseded by https://github.com/trikoder/oauth2-bundle/pull/167

florianajir:
no problem 👍

PR #118, opened 2019-11-23
Symfony 5.0

rbaarsma:
Adds Symfony 5 support and bumps minimum symfony version to 4.4 (as of January symfony 4.3 is no longer actively maintained, only security fixes and in the master of this bundle I opt to keep up with the times. If anyone uses an older version they will automatically use an older version of this bundle)

PR #141, opened 2019-12-07
Update PHP-CS-Fixer to support PHP 7.4

X-Coder264:


X-Coder264:
https://github.com/trikoder/oauth2-bundle/pull/137 was not enough. Even though we no longer depend on `symfony/contracts`, `symfony/symfony` 4.2 still does so we are running into the same problem. Since dropping SF 4.2 in the CI is IMO not an option, we will have to wait on https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4685 to be merged and tagged before we can proceed with this PR.

rbaarsma:
@X-Coder264 I've noticed this issue is already open for 2 weeks, because the related issue is still open. And my PR https://github.com/trikoder/oauth2-bundle/pull/118 is also waiting on this.

I'm wondering why you don't want to drop symfony 4.2 support? It's no longer maintained. As I've done in my PR, the minimum symfony version for newer versions of this bundle should, in my opinion, be 4.4. Not sure if that fixes the issue?

X-Coder264:
@rbaarsma I've pinged the maintainers in the PHP-CS fixer repo to see what is the status on merging the fix. Hopefully it's gonna be merged and tagged soon. If not we will explore other options.

As far as your question regarding Symfony 4.2 - we don't like dropping support for Symfony versions in minor releases (unless there is a significant maintenance burden which isn't the case with SF 4.2). A lot of people are still using 4.2 for whatever reasons and from our POV there is no need to force them to upgrade if they want to keep getting the bug fixes and new stuff that gets merged in the 2.x branch.

rbaarsma:
@X-Coder264 seems the story where this PR was waiting for is merged. Perhaps this can be merged too and then the php 7.4 and symfony 5 story as well?

rbaarsma:
Let me know where I can help

X-Coder264:
@rbaarsma Yes, the CS-fixer PR got merged, but they haven't tagged a new version with the fix yet. As soon as they do that (hopefully in the next couple of days) we will merge this and let you know so that you can finish your PR.

PR #167, opened 2020-02-20
Add full support for public clients for auth code grant (PKCE)

X-Coder264:


Tests/Unit/ClientEntityTest.php
```diff
@@ -0,0 +1,30 @@
...
+        $client = new Client('foo', $secret);
+
+        $this->assertSame($isConfidential, $client->isConfidential());
+    }
+
+    public function confidentialDataProvider(): iterable
+    {
+        return [
+            'Client with null secret is not confidential' => [null, false],
+            'Client with empty secret is not confidential' => ['', false],
```
HypeMC: @X-Coder264 Why do we want this case?

Command/CreateClientCommand.php
```diff
@@ -67,13 +69,28 @@ protected function configure(): void
                 InputArgument::OPTIONAL,
                 'The client secret'
             )
+            ->addOption(
+                'public',
+                null,
+                InputOption::VALUE_OPTIONAL,
```
HypeMC: @X-Coder264 Is this suppose to have a value? If it's not a bool flag then how is it going to be used, eg
```
--public=0
--public=1
```
If that's the case I wouldn't add `false` as a default since it's not consistent. Maybe `'0'`?

Command/CreateClientCommand.php
```diff
@@ -89,7 +106,14 @@ protected function execute(InputInterface $input, OutputInterface $output): int
     private function buildClientFromInput(InputInterface $input): Client
     {
         $identifier = $input->getArgument('identifier') ?? hash('md5', random_bytes(16));
-        $secret = $input->getArgument('secret') ?? hash('sha512', random_bytes(32));
+
+        $isPublic = $input->getOption('public');
```
HypeMC: @X-Coder264 Also, if this is not a bool flag, then shouldn't we cast to bool here?

Command/CreateClientCommand.php
```diff
@@ -67,13 +69,28 @@ protected function configure(): void
                 InputArgument::OPTIONAL,
                 'The client secret'
             )
+            ->addOption(
+                'public',
+                null,
```
HypeMC: @X-Coder264 This seems like a thing that could be used often, maybe we should add a shortcut?

Tests/Acceptance/CreateClientCommandTest.php
```diff
@@ -35,11 +35,66 @@ public function testCreateClientWithIdentifier(): void
...
+
+    public function testCreatePublicClientWithIdentifier(): void
+    {
+        $clientIdentifier = 'foobar test';
+        $command = $this->application->find('trikoder:oauth2:create-client');
+        $commandTester = new CommandTester($command);
+        $commandTester->execute([
+            'command' => $command->getName(),
+            'identifier' => $clientIdentifier,
+            '--public' => true,
```
HypeMC: @X-Coder264 Same here, is it bool?

Tests/Unit/ClientEntityTest.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Trikoder\Bundle\OAuth2Bundle\Tests\Unit;
+
+use PHPUnit\Framework\TestCase;
+use Trikoder\Bundle\OAuth2Bundle\Model\Client;
+
+class ClientEntityTest extends TestCase
```
HypeMC: @X-Coder264 `final`

Model/Client.php
```diff
@@ -113,4 +113,9 @@ public function setActive(bool $active): self
 
         return $this;
     }
+
+    public function isConfidential(): bool
+    {
+        return !empty($this->secret);
```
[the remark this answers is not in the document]
X-Coder264: @HypeMC Yes, that was the exact intention here as evidenced by the unit test I've written for this.

Tests/Unit/ClientEntityTest.php
X-Coder264: @HypeMC Passport also behaves in the same way: https://github.com/laravel/passport/blob/v8.4.0/src/Client.php#L103

Model/Client.php
```diff
@@ -36,7 +36,7 @@ class Client
      */
     private $active = true;
 
-    public function __construct(string $identifier, string $secret)
+    public function __construct(string $identifier, ?string $secret)
```
HypeMC: @X-Coder264 Should we add a default null value here?

X-Coder264: @HypeMC I don't think there is a need for that. The secret was a required argument even before this PR so we are explicitly passing the argument in all code paths.

Tests/Acceptance/TokenEndpointTest.php
```diff
@@ -157,6 +157,38 @@ public function testSuccessfulAuthorizationCodeRequest(): void
         $this->assertNotEmpty($jsonResponse['access_token']);
     }
 
+    public function testSuccessfulAuthorizationCodeRequestWithPublicClient(): void
```
spideyfusion: @X-Coder264 Could you please add a test case for this code path?

https://github.com/thephpleague/oauth2-server/blob/8.0.0/src/Grant/AuthCodeGrant.php#L316

Command/CreateClientCommand.php
```diff
@@ -67,13 +69,28 @@ protected function configure(): void
...
     }
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $io = new SymfonyStyle($input, $output);
-        $client = $this->buildClientFromInput($input);
+
+        try {
+            $client = $this->buildClientFromInput($input);
+        } catch (Exception $exception) {
```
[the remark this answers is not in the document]
X-Coder264: @HypeMC We don't. Fixed.

Command/CreateClientCommand.php
X-Coder264: @HypeMC It's a bool flag.

Command/CreateClientCommand.php
X-Coder264: @HypeMC It's a bool flag.

Tests/Acceptance/CreateClientCommandTest.php
X-Coder264: @HypeMC It's a bool flag.

Tests/Unit/ClientEntityTest.php
X-Coder264: @HypeMC Added.

Command/CreateClientCommand.php
X-Coder264: @HypeMC We don't provide shortcuts for any of the options in this command and I don't think we should provide a shortcut for creating a public client - it's always best to be explicit when dealing with such security related stuff.

Tests/Acceptance/TokenEndpointTest.php
X-Coder264: @spideyfusion Sure. I added the test.

Command/CreateClientCommand.php
```diff
@@ -67,13 +68,27 @@ protected function configure(): void
                 InputArgument::OPTIONAL,
                 'The client secret'
             )
+            ->addOption(
+                'public',
+                null,
+                InputOption::VALUE_NONE,
+                'Create a public client'
```
HypeMC: @X-Coder264 OMG, just noticed there's no dot at the end here while all the other options have one. Cannot... unsee... ocd... to strong...

X-Coder264: 😂

X-Coder264: But all arguments have no dot at the end, what a major inconsistency 😱

HypeMC: I know, this is killing me now...

X-Coder264: 🤣 😂

spideyfusion: What is killing me is this unresolved discussion... 😂

PR #168, opened 2020-02-21
Make the token aware of the firewall context

X-Coder264:
In the scenario where the application has multiple firewall contexts defined in their security layer when an issued token gets passed to the `AuthenticationProviderManager`, the token will get processed by the first registered authentication provider that can handle the token. This is a problem when the application has firewall contexts with different user providers.

### DIFF
--- a/Tests/Unit/OAuth2TokenTest.php
+++ b/Tests/Unit/OAuth2TokenTest.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Trikoder\Bundle\OAuth2Bundle\Tests\Unit;
+
+use PHPUnit\Framework\TestCase;
+use Psr\Http\Message\ServerRequestInterface;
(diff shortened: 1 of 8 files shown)


Tests/Unit/OAuth2TokenTest.php
```diff
@@ -0,0 +1,46 @@
...
+
+        $roles = $unserializedToken->getRoles();
+        $this->assertCount(1, $roles);
+        $expectedRole = $rolePrefix . strtoupper($scopes[0]);
+        $this->assertSame($expectedRole, $roles[0]->getRole());
+
+        $this->assertSame($user->getUsername(), $unserializedToken->getUser()->getUsername());
+        $this->assertFalse($unserializedToken->isAuthenticated());
+
+        if (method_exists($token, 'getRoleNames')) {
```
HypeMC: @X-Coder264 This check wont be necessary when we dorm Sf 3.4 support, so I'd add a comment here as well.

PR #171, opened 2020-02-23
Update debugging documentation

X-Coder264:
In https://github.com/trikoder/oauth2-bundle/pull/141 we forgot to update the debugging section in the README.md. As timecop function overriding is now disabled by default in our Docker image running this line (without the change in this PR) would result in a bunch of tests failing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ dev/bin/php composer test
 You can run the debugger using the following command:
 
 ```sh
-dev/bin/php-debug vendor/bin/phpunit
+dev/bin/php-debug -d timecop.func_override=1 vendor/bin/phpunit
 ```
 
 Make sure your IDE is setup properly, for more information check out the [dedicated documentation](docs/debugging.md).


README.md
```diff
@@ -183,7 +183,7 @@ dev/bin/php composer test
 You can run the debugger using the following command:
 
 ```sh
-dev/bin/php-debug vendor/bin/phpunit
+dev/bin/php-debug -d timecop.func_override=1 vendor/bin/phpunit
```
HypeMC: @X-Coder264 I don't like this approach at all. I think that the original PR #141 should't have added docker specific stuff (the `php -d timecop.func_override=1` part) to `composer.json` in the first place. A better solution would be to add a helper script, eg `dev/bin/php` to run tests & add the `-d timecop.func_override=1` part to `php-debug` as well for debugging.

spideyfusion: @HypeMC Let's continue the new approach in https://github.com/trikoder/oauth2-bundle/pull/173 and merge this documentation update just so we don't provide people with broken instructions.

PR #177, opened 2020-02-25
[WIP] Add documentation for authorization code grant 

### LINKED ISSUE #88: Permission request page
When authentication is using the grant type authorization_code, the user should be redirected after loging in succesfully to a page where permissions are asked to the user.

If the user allows the application using his login data, he should be redirected to the original application where he will be logged in to. If he doesn't give permissions, the user should be redirected to the application without being logged in.

### LINKED ISSUE #160: Document how to setup the "authorization_code" grant

v-m-i:
Closes #160

### DIFF
--- a/docs/authorization-code-grant.md
+++ b/docs/authorization-code-grant.md
@@ -1,0 +1,124 @@
+# Authorization code grant
+
+Authorization code grant has two steps
+
+1. Acquiring authorization code
+2. Getting token from authorization code
+
+## Requirements
+
+To use authorization code grant `enable_auth_code_grant` parameter inside `authorization_server` must be set to `true` (it is set to `true` by default).
+
+### Example: config.yml
+
+```yaml
+trikoder_oauth2:
+    authorization_server:
+        enable_auth_code_grant: true
+```
+
+After authorization code grant is enabled, token and authorization endpoints must be set.
+It can be done by including `Resources/config/routes.xml` which will provide `/authorize` or `/token` endpoints or manually by setting
+
+1. Controller `Trikoder\Bundle\OAuth2Bundle\Controller\AuthorizationController::indexAction` with `GET` method for authorization endpoint
+2. Controller `Trikoder\Bundle\OAuth2Bundle\Controller\TokenController::indexAction` with `POST` method for token endpoint
+
+### Example: custom setup
+
+```yaml
+oauth2_authorization_code:
+    controller: Trikoder\Bundle\OAuth2Bundle\Controller\AuthorizationController::indexAction
+    path: /oauth2-authorization-code
+
+oauth2_token:
+    controller: Trikoder\Bundle\OAuth2Bundle\Controller\TokenController::indexAction
+    path: /api/token
+```
+
+After assigning routes, listener for `trikoder.oauth2.authorization_request_resolve` must be configured.
+
+`\Trikoder\Bundle\OAuth2Bundle\Event\AuthorizationRequestResolveEvent` (whose name is `trikoder.oauth2.authorization_request_resolve`) consist of three important methods which have to be used
+
+1. `setUser(?UserInterface $user)` and `resolveAuthorization(bool $authorizationResolution)` when user is already logged in when accessing authorization endpoint
+2. `setResponse(ResponseInterface $response)` when user needs to log in before authorization server can issue authorization code
+
+### Example: (services.yml and php class)
+
+```yaml
+    BestNamespace\OAuthLogin\Listener\AuthorizationCodeListener:
+        tags:
+            - { name: kernel.event_listener, event: 'trikoder.oauth2.authorization_request_resolve', method: onAuthorizationRequestResolve }
+```
+
+```php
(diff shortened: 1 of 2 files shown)


okazy (commented):
@v-m-i 
Thanks for the great documentation.
I didn't understand the following two points, so the documentation was very helpful.

- How to change the path of the endpoint
- How to handle redirect before authentication

I commented on some interesting points.

I have implemented my own authorization confirmation page.

1. login page
1. authorization confirmation page(ex. Do you allow read permission for this application?)
1. redirect with authorization code

I would like to know if there is a best practice on how to set the authorization confirmation page.

Here is my code implemented without documentation.
https://github.com/okazy/ec-cube/compare/f604226b42131689f763a2d87af0994173c87390...okazy:d981ba70858282c1353f36367299db04053d03ff?expand=1

docs/authorization-code-grant.md
```diff
@@ -0,0 +1,124 @@
...
+        RequestStack $requestStack
+    ) {
+        $this->security = $security;
+        $this->urlGenerator = $urlGenerator;
+        $this->requestStack = $requestStack;
+    }
+
+    public function onAuthorizationRequestResolve(AuthorizationRequestResolveEvent $event)
+    {
+        if (null !== ($user = $this->security->getUser())) {
```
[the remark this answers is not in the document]
v-m-i: @okazy There is no difference, I have changed that part and added notice on that, good catch :+1:

v-m-i:
@okazy 

Thank you for your feedback, I have committed your suggestions and added some additional changes for listener example.

Regarding authorization confirmation page, I don't know best practice for creating it. 

@HypeMC 
What do you think, maybe open new issue for best practice of creating authorization confirmation page? Also, maybe we should open issue (or include in best practice issue) explaining what is best practice for [restricting scopes on authorization server](https://tools.ietf.org/html/rfc6749#section-3.3)? (I don't have that use-case in my applications so I can't recommend anything)

docs/authorization-code-grant.md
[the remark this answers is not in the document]
okazy: Thanks, I got it :)

mnsami:
@v-m-i 

Thank you for this documentation it helped me with my project a lot.

I have a question, in the documentation it is mentioned to use `setResponse` when we want the user to login before issuing the Authorization code, but for me it is not yet clear to me how.

Currently, I have a project where I'm using `nucleos/user-bundle` for user management, login, in an API project and I hooked up `trikoder/oauth2-bundle` to issue a client for my `front-end` reactjs app.

I want to use the login mechanism along side the Authorization code to make the user via my reactjs app make API calls, but I want to force the login like you mentioned in the documentation.

So, can you explain more about the `setResponse` please? thanks.

docs/authorization-code-grant.md
```diff
@@ -0,0 +1,127 @@
+# Authorization code grant
+
+Authorization code grant has two steps
+
+1. Getting authorization code
+2. Getting token from authorization code
+
+## Requirements
+
+To use authorization code grant `enable_auth_code_grant` parameter inside `authorization_server` must be set to `true` (it is set to `true` by default).
```
tdutrion, line 10:
```
[info] User Deprecated: "trikoder_oauth2.authorization_server.enable_auth_code_grant" is deprecated, use "trikoder_oauth2.authorization_server.grant_types.authorization_code.enable" instead.
```

This should be updated to use the new parameters.

docs/authorization-code-grant.md
```diff
@@ -0,0 +1,127 @@
...
+        UrlGeneratorInterface $urlGenerator,
+        RequestStack $requestStack
+    ) {
+        $this->urlGenerator = $urlGenerator;
+        $this->requestStack = $requestStack;
+    }
+
+    public function onAuthorizationRequestResolve(AuthorizationRequestResolveEvent $event)
+    {
+        if (null !== $event->getUser()) {
```
tdutrion, line 86: You could reverse the condition (`null === $event->getUser()`) and just return after setting the response:

```php
public function onAuthorizationRequestResolve(AuthorizationRequestResolveEvent $event): void
{
    if (null === $event->getUser()) {
        $event->setResponse(new Response(302, [
            'Location' => $this->urlGenerator->generate('login', [
                'returnUrl' => $this->requestStack->getMasterRequest()->getUri(),
            ]),
        ]));

        return;
    }

    $event->resolveAuthorization(AuthorizationRequestResolveEvent::AUTHORIZATION_APPROVED);
}
```

tdutrion (commented):
Hi! Thanks for this PR, it has been very useful to me!

Just a couple of changes that might be interesting here :)

docs/authorization-code-grant.md
v-m-i: @tdutrion
I agree, simplifying if statements using early returns is (usually) good practice. I have updated the documentation with your example.

smilesrg:
@trikoder maybe it's time to accept this PR?

Issue #143, opened 2019-12-11
Add custom grant types


ping86:
Sometimes you need add custom types:

Related issue: https://github.com/thephpleague/oauth2-server/issues/611

Now we add more grant types with compiler pass:

```
    private function configureSocialGrantType(ContainerBuilder $container): void
    {
        $definition = $container->findDefinition(AuthorizationServer::class);

        $definition->addMethodCall('enableGrantType', [
            new Reference(SocialGrant::class)
        ]);
    }
```

We need add a custom grant type for example for social login with google. But the model avoid use custom grant types:

```
        if (!OAuth2Grants::has($grant)) {
            throw new RuntimeException(sprintf('The \'%s\' grant is not supported.', $grant));
        }
```
https://github.com/trikoder/oauth2-bundle/blob/v2.x/Model/Grant.php#L20

```
    public static function has(string $grant): bool
    {
        return \in_array($grant, [
            self::CLIENT_CREDENTIALS,
            self::PASSWORD,
            self::REFRESH_TOKEN,
            self::AUTHORIZATION_CODE,
            self::IMPLICIT,
        ]);
    }
```
https://github.com/trikoder/oauth2-bundle/blob/v2.x/OAuth2Grants.php#L44

It would be fantastic can enable custom grant types with bundle configuration.

Thanks.

spideyfusion:
We added support for registering custom grant types via https://github.com/trikoder/oauth2-bundle/pull/178.

PR #178, opened 2020-02-25
Add support for registering custom grant types

X-Coder264:
Closes https://github.com/trikoder/oauth2-bundle/issues/143


DependencyInjection/TrikoderOAuth2Extension.php
```diff
@@ -70,6 +71,9 @@ public function load(array $configs, ContainerBuilder $container)
                 'method' => 'onKernelException',
                 'priority' => $config['exception_event_listener_priority'],
             ]);
+
+        $container->registerForAutoconfiguration(GrantTypeInterface::class)
+            ->addTag('authorization_server.grant');
```
HypeMC, line 76: @X-Coder264 @spideyfusion I think we should prefix this with `trikoder.oauth2.`, WDYT?

spideyfusion: @HypeMC @X-Coder264 Agreed.

PR #187, opened 2020-03-24
Revoking credentials

### LINKED ISSUE #182: Mark internal classes
If I'm not mistaken, the library doesn't expose public API at the moment.

I think it makes sense to mark all classes that are not meant to be used as [`@internal`](https://docs.phpdoc.org/latest/references/phpdoc/tags/internal.html).

With that said, any class that doesn't have this mark should be treated as public API of the package which would help with easier versioning and development, as breaking API of internal classes would not be considered a breaking change of the package itself that should yield a major version bump.

toniperic:
Hi guys.

I've gotten a case where I want to invalidate all currently issued credentials per client or user level. And by "credentials" I mean access tokens, refresh tokens and authorization codes.

A real life example where one could use this is if a user changes their account password - you might immediately ask the user whether they would like to force a logout on all the devices they were currently signed in.

This PR adds a public API so this can be done programmatically out-of-the-box.

Looking forward to getting some feedback.

Cheers!

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -4,6 +4,8 @@
            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
+        <service id="Trikoder\Bundle\OAuth2Bundle\Service\ClientFinderInterface" alias="Trikoder\Bundle\OAuth2Bundle\Manager\ClientManagerInterface" />
+
         <!-- The league repositories -->
         <service id="Trikoder\Bundle\OAuth2Bundle\League\Repository\ClientRepository">
             <argument type="service" id="Trikoder\Bundle\OAuth2Bundle\Manager\ClientManagerInterface" />
--- a/Service/ClientFinderInterface.php
+++ b/Service/ClientFinderInterface.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Trikoder\Bundle\OAuth2Bundle\Service;
+
+use Trikoder\Bundle\OAuth2Bundle\Model\Client;
+
+/**
+ * @api
+ */
+interface ClientFinderInterface
+{
+    public function find(string $identifier): ?Client;
+}
--- a/Service/CredentialsRevoker/DoctrineCredentialsRevoker.php
+++ b/Service/CredentialsRevoker/DoctrineCredentialsRevoker.php
@@ -1,0 +1,101 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Trikoder\Bundle\OAuth2Bundle\Service\CredentialsRevoker;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Symfony\Component\Security\Core\User\UserInterface;
+use Trikoder\Bundle\OAuth2Bundle\Model\AccessToken;
+use Trikoder\Bundle\OAuth2Bundle\Model\AuthorizationCode;
+use Trikoder\Bundle\OAuth2Bundle\Model\Client;
+use Trikoder\Bundle\OAuth2Bundle\Model\RefreshToken;
+use Trikoder\Bundle\OAuth2Bundle\Service\CredentialsRevokerInterface;
+
+final class DoctrineCredentialsRevoker implements CredentialsRevokerInterface
+{
+    /**
+     * @var EntityManagerInterface
+     */
+    private $entityManager;
+
+    public function __construct(EntityManagerInterface $entityManager)
+    {
+        $this->entityManager = $entityManager;
+    }
+
+    public function revokeCredentialsForUser(UserInterface $user): void
+    {
+        $userIdentifier = $user->getUsername();
+
+        $this->entityManager->createQueryBuilder()
+            ->update(AccessToken::class, 'at')
+            ->set('at.revoked', true)
+            ->where('at.userIdentifier = :userIdentifier')
+            ->setParameter('userIdentifier', $userIdentifier)
+            ->getQuery()
+            ->execute();
+
+        $queryBuilder = $this->entityManager->createQueryBuilder();
+        $queryBuilder
+            ->update(RefreshToken::class, 'rt')
+            ->set('rt.revoked', true)
+            ->where($queryBuilder->expr()->in(
+                'rt.accessToken',
+                $this->entityManager->createQueryBuilder()
+                    ->select('at.identifier')
+                    ->from(AccessToken::class, 'at')
+                    ->where('at.userIdentifier = :userIdentifier')
+                    ->getDQL()
+            ))
+            ->setParameter('userIdentifier', $userIdentifier)
+            ->getQuery()
+            ->execute();
+
+        $this->entityManager->createQueryBuilder()
+            ->update(AuthorizationCode::class, 'ac')
+            ->set('ac.revoked', true)
+            ->where('ac.userIdentifier = :userIdentifier')
+            ->setParameter('userIdentifier', $userIdentifier)
+            ->getQuery()
+            ->execute();
+    }
+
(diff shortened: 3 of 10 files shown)


spideyfusion:
@toniperic Please separate this new feature addition from public API defining into a new PR. All PRs should target `v3.x` unless they're introducing a breaking change for consumers.

Service/ClientFinderInterface.php
```diff
@@ -0,0 +1,13 @@
...
+
+declare(strict_types=1);
+
+namespace Trikoder\Bundle\OAuth2Bundle\Service;
+
+use Trikoder\Bundle\OAuth2Bundle\Model\Client;
+
+interface ClientFinderInterface
+{
+    /** @api */
```
spideyfusion: @toniperic Use the full docblock format when describing structural elements. For example:

```php
/**
 * @api
 */
```

Service/ClientFinderInterface.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Trikoder\Bundle\OAuth2Bundle\Service;
+
+use Trikoder\Bundle\OAuth2Bundle\Model\Client;
+
+interface ClientFinderInterface
```
spideyfusion: @toniperic Since we know this is going to be apart of our public API, please add comments in the class header that describe the interface's purpose (same goes for the method itself).

Also, can we just put `@api` on the entire interface instead of an individual method?

Service/CredentialsRevokerInterface.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Trikoder\Bundle\OAuth2Bundle\Service;
+
+use Symfony\Component\Security\Core\User\UserInterface;
+use Trikoder\Bundle\OAuth2Bundle\Model\Client;
+
+interface CredentialsRevokerInterface
```
spideyfusion: @toniperic Since we know this is going to be apart of our public API, please add comments in the class header that describe the interface's purpose (same goes for the method itself).

Service/CredentialsRevoker/DoctrineRevoker.php
```diff
@@ -0,0 +1,99 @@
...
+use Symfony\Component\Security\Core\User\UserInterface;
+use Trikoder\Bundle\OAuth2Bundle\Model\AccessToken;
+use Trikoder\Bundle\OAuth2Bundle\Model\AuthorizationCode;
+use Trikoder\Bundle\OAuth2Bundle\Model\Client;
+use Trikoder\Bundle\OAuth2Bundle\Model\RefreshToken;
+use Trikoder\Bundle\OAuth2Bundle\Service\CredentialsRevokerInterface;
+
+final class DoctrineRevoker implements CredentialsRevokerInterface
+{
+    private $entityManager;
```
spideyfusion: @toniperic You're missing a docblock for this property.

Service/CredentialsRevoker/DoctrineRevoker.php
```diff
@@ -0,0 +1,99 @@
...
+
+use Doctrine\ORM\EntityManagerInterface;
+use Symfony\Component\Security\Core\User\UserInterface;
+use Trikoder\Bundle\OAuth2Bundle\Model\AccessToken;
+use Trikoder\Bundle\OAuth2Bundle\Model\AuthorizationCode;
+use Trikoder\Bundle\OAuth2Bundle\Model\Client;
+use Trikoder\Bundle\OAuth2Bundle\Model\RefreshToken;
+use Trikoder\Bundle\OAuth2Bundle\Service\CredentialsRevokerInterface;
+
+final class DoctrineRevoker implements CredentialsRevokerInterface
```
spideyfusion: @toniperic Should be named `DoctrineCredentialsRevoker`.

Service/CredentialsRevoker/DoctrineRevoker.php
```diff
@@ -0,0 +1,99 @@
...
+
+    public function revokeCredentialsForUser(UserInterface $user): void
+    {
+        $userIdentifier = $user->getUsername();
+
+        $this->entityManager->createQueryBuilder()
+            ->update(AccessToken::class, 'at')
+            ->set('at.revoked', true)
+            ->where('at.userIdentifier = :userIdentifier')
+            ->setParameter('userIdentifier', $userIdentifier)
```
spideyfusion: @toniperic Also, add an index for the `userIdentifier` field so we get a nice performance boost when revoking tokens.

Service/ClientFinderInterface.php
toniperic: Yeah, moved `@api` to the interface.

However I'm not sure how I'd additionally describe such contract, could you maybe help?

It finds clients, as the contract suggests, so adding comments without actually providing any additional information seems a bit redundant to me?

toniperic:
Hey @spideyfusion, I've addressed all of the comments.

Looking forward to your feedback.

Resources/config/services.xml
```diff
@@ -4,6 +4,8 @@
            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
+        <service id="Trikoder\Bundle\OAuth2Bundle\Service\ClientFinderInterface" alias="Trikoder\Bundle\OAuth2Bundle\Manager\ClientManagerInterface" />
```
HypeMC: @toniperic This doesn't belong here IMO, I'd put it in `Resources/config/storage/doctrine.xml`.
@spideyfusion  WDYT?

Service/CredentialsRevoker/DoctrineCredentialsRevoker.php
```diff
@@ -0,0 +1,102 @@
...
+            ->set('at.revoked', true)
+            ->where('at.client = :client')
+            ->setParameter('client', $doctrineClient)
+            ->getQuery()
+            ->execute();
+
+        $qb = $this->entityManager->createQueryBuilder();
+        $qb->update(RefreshToken::class, 'rt')
+            ->set('rt.revoked', true)
+            ->where($qb->expr()->in(
```
HypeMC: @toniperic Could we perhaps use a join here instead of sub-queries?

toniperic: There are two kinda-connected reasons why I've opted for doing a sub-query instead of a join:

1. [This](https://stackoverflow.com/a/2577188/7424875)
2. Some storage drivers that Doctrine supports do not support UPDATEs with JOINs _(eyes at SQLite, but there may be more...)_

HypeMC: @toniperic Ok, makes sense.

Service/CredentialsRevoker/DoctrineCredentialsRevoker.php
```diff
@@ -0,0 +1,102 @@
...
+            ->setParameter('userIdentifier', $userIdentifier)
+            ->getQuery()
+            ->execute();
+
+        $queryBuilder = $this->entityManager->createQueryBuilder();
+        $queryBuilder
+            ->update(RefreshToken::class, 'rt')
+            ->set('rt.revoked', true)
+            ->where(
+                $queryBuilder->expr()->in(
```
HypeMC, line 44: @toniperic A little consistency issue:
```suggestion
            ->where($queryBuilder->expr()->in(
```

Resources/config/services.xml
toniperic: @HypeMC I think it belongs here due to the fact that if this was in `doctrine.xml` and someone configured the bundle to use the in-memory persistence method, then this alias wouldn't be registered at all.

HypeMC: That a very good point, nice catch.

spideyfusion:
@toniperic Thank you for your contribution! :confetti_ball:

toniperic:
You're welcome :)

PR #192, opened 2020-03-30
Fix UpdateClientCommand options

### LINKED ISSUE #191: UpdateClientCommand options behavior
I have noticed that UpdateClientCommand deletes options that were not sent and always activates client if `deactivated` flag was not sent.

Example:

```
bin/console trikoder:oauth2:create-client client test --grant-type client_credentials
bin/console trikoder:oauth2:update-client client --deactivated 
 => client now has active set to 0 and grants set to null (lost his client_credentials grant)
bin/console trikoder:oauth2:update-client client --scope test
 => client now has active set to 1, grants to null and scope to test
```

Expected end result for example above would be: 
Client has active set to 0, grants to client_credentials and scope to test.

v-m-i:
Closes #191 

Array options won't be deleted if they were not sent.
Client won't be activated if `deactivated` flag was not sent. 
Removed `deactivated`, added `active` option.

### DIFF
--- a/Command/UpdateClientCommand.php
+++ b/Command/UpdateClientCommand.php
@@ -39,29 +39,30 @@ final class UpdateClientCommand extends Command
             ->addOption(
                 'redirect-uri',
                 null,
-                InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY,
-                'Sets redirect uri for client. Use this option multiple times to set multiple redirect URIs.',
-                []
+                InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
+                'Sets redirect uri for client. Use this option multiple times to set multiple redirect URIs. Use it without value to remove existing values.',
+                [0]
             )
             ->addOption(
                 'grant-type',
                 null,
-                InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY,
-                'Sets allowed grant type for client. Use this option multiple times to set multiple grant types.',
-                []
+                InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
+                'Sets allowed grant type for client. Use this option multiple times to set multiple grant types. Use it without value to remove existing values.',
+                [0]
             )
             ->addOption(
                 'scope',
                 null,
-                InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY,
-                'Sets allowed scope for client. Use this option multiple times to set multiple scopes.',
-                []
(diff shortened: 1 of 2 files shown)


Command/UpdateClientCommand.php
```diff
@@ -39,29 +39,30 @@ protected function configure(): void
             ->addOption(
                 'redirect-uri',
                 null,
-                InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY,
-                'Sets redirect uri for client. Use this option multiple times to set multiple redirect URIs.',
-                []
+                InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
+                'Sets redirect uri for client. Use this option multiple times to set multiple redirect URIs. Use it without value to remove existing values.',
+                [0]
```
HypeMC: @v-m-i I have several issues with this approach:
1) this hole `[0]` approach seems kind of hacky, why not just leave `[]` as the default & use `empty`:
image
2) what if someone runs it like this: `bin/console trikoder:oauth2:update-client 1 --redirect-uri=url --redirect-uri`, they would get:
image

I'm thinking maybe this would be a better approach:
1) leave `InputOption::VALUE_REQUIRED` & `[]` as the default
2) add `--remove-*` equivalents, eg `--remove-grant-type`

Then if someone has a client with the scopes `s1` & `s2`, they could do: `trikoder:oauth2:update-client 1 --scope=s3 --remove-scope=s1` & would end up with a client that has the scopes `s2` & `s3`. Hint: you could use `array_merge` & `array_diff`.

@spideyfusion @X-Coder264 Any thoughts on this one?

HypeMC (changes requested):

Command/UpdateClientCommand.php
v-m-i: Hi @HypeMC 

1. I agree it is hacky, you are right, there is no reason to use `[0]` when I could have used `[]`, I will change it if we chose this "set" approach.

2. This comand sets attributes it has recieved, meaning that after running `trikoder:oauth2:update-client 1 --scope=s3` client `1` would have only `s3` scope, regardless of how many scopes it had before. 

We have two requirements to satisfy here:
    1. User can remove all scopes from client
    2. User can set scopes for client

I think that in order to satisfy first requirement our InputOption must be `VALUE_OPTIONAL` because `trikoder:oauth2:update-client 1 --scope` is the only way to remove all scopes with "set" approach?

Add/Remove approach you have mentioned is something I would like more than this "set" approach. If we would decide to go with that approach, I would suggest that `--scope`, and other methods are prefixed with `add`, like this: `trikoder:oauth2:update-client 1 --add-scope=s3 --remove-scope=s1`

PR #199, opened 2020-04-11
Add per grant type configuration options

### LINKED ISSUE #105: change access_token_ttl only for client_credentials
Is it possible to change the access_token_ttl only for client_credentials? I wanted to have different expirations for client_credentials and for password.

Tks

HypeMC:
Resolves #105.


GordonGuenther:
Would this PR also allow to enable and configure custom grant types?

X-Coder264:
@GordonGuenther You can already enable and configure custom grant types without this PR -> https://github.com/trikoder/oauth2-bundle/blob/v3.x/docs/implementing-custom-grant-type.md

PR #208, opened 2020-04-20
Dev improvements

spideyfusion:


HypeMC:
@spideyfusion Shouldn't this go on v3.x? As far as I'm concerned these types of fixes can go on v2.x in case we need to do some bug fixes on that version since we are still supporting it.

PR #215, opened 2020-07-24
Custom entities

### LINKED ISSUE #145: Support for custom entities
Hi! 

Is there a way to change the built-in entites from the bundle to custom ones? The reason behind this is to create some additional fields that would implement given business logic (some clients are public, some are default, some are premoderated etc.)

ricardodevries:
Closes #145

Add the ability to use custom entities. This unfortunately means the user has to create the bundle entities themselves in order to use the bundle. A small price to pay for customization, imo.

### DIFF
--- a/Tests/TestKernel.php
+++ b/Tests/TestKernel.php
@@ -141,7 +141,18 @@ final class TestKernel extends Kernel implements CompilerPassInterface
                 ],
                 'platform_service' => SqlitePlatform::class,
             ],
-            'orm' => null,
+            'orm' => [
+                'auto_mapping' => true,
+                'mappings' => [
+                    'App' => [
+                        'type' => 'annotation',
+                        'dir' => '../docs/resources/entities',
+                        'is_bundle' => false,
+                        'prefix' => 'App\Entity',
+                        'alias' => 'App',
+                    ],
+                ],
+            ],
         ]);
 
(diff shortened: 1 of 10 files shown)


ricardodevries:
PR in draft mode until tests are fixed.

ricardodevries:
Perhaps another contributor or maintainer could figure out what the correct doctrine mapping source is in the project.

Tests/TestKernel.php
```diff
@@ -141,7 +141,18 @@ protected function configureContainer(ContainerBuilder $container, LoaderInterfa
                 ],
                 'platform_service' => SqlitePlatform::class,
             ],
-            'orm' => null,
+            'orm' => [
+                'auto_mapping' => true,
+                'mappings' => [
+                    'App' => [
+                        'type' => 'annotation',
+                        'dir' => '../docs/resources/entities',
```
ricardodevries, line 149: Perhaps another contributor or maintainer could figure out what the correct doctrine mapping source is in the project?

ricardodevries, line 149: @HypeMC or @X-Coder264; In order to fix the tests, perhaps one of you could figure out what the correct doctrine mapping source is in the project to the demo entities I created in `docs/resources/entities`? It seems useless to re-create these files just for testing.

I would like to have your opinion on this matter.

PR #216, opened 2020-07-28
Set "CASCADE" on delete for AuthorizationCode/Client association

yceruto:
Same as `AccessToken` model:
https://github.com/trikoder/oauth2-bundle/blob/6ea74bcef303f1a0b70975336ac0878c381cc21e/Resources/config/doctrine/model/AccessToken.orm.xml#L18

if the a Client row is being deleted, then the related authorization codes should be delete too.


X-Coder264:
Thanks @yceruto for the PR. Will add a note to the upgrade section when we tag the next release so that people see that the mapping was updated and that they should update their DB.

yceruto:
Thank you @X-Coder264 for your quick response and merging.

X-Coder264:
Np, thank you for the contribution.

PR #219, opened 2020-07-30
Update README.md

Orkin:
Add symfony-bundles/json-request-bundle package to parse json content on oauth request. By default json content is handle as string and did not work with oauth2 request json content

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ This package is currently in the active development.
 1. Require the bundle and a PSR 7/17 implementation with Composer:
 
     ```sh
-    composer require trikoder/oauth2-bundle nyholm/psr7
+    composer require trikoder/oauth2-bundle nyholm/psr7 symfony-bundles/json-request-bundle
     ```
 
     If your project is managed using [Symfony Flex](https://github.com/symfony/flex), the rest of the steps are not required. Just follow the post-installation instructions instead! :tada:


X-Coder264:
Sorry, but that package is not required in order to use this bundle. We could maybe evaluate if it'd be helpful to mention it in the composer.json `suggest` section if you can provide a good reason for that as I'm not sure that I completely understood what you've written.

Orkin:
@X-Coder264 I agree that is not required to use this bundle but if you do an oauth request as json this bundle is not working properly so I think it can be useful to mention that somewhere.

X-Coder264:
What does "not working properly" mean exactly? Please give an example.

Orkin:
@X-Coder264

```
POST /oauth/token HTTP/1.1
Content-Type: application/json
Accept: application/json
Accept-Language: fr-FR
Host: localhost:8002
Connection: close
User-Agent: Paw/3.1.10 (Macintosh; OS X/10.15.6) GCDHTTPRequest
Content-Length: 282

{"username":"a-user-name","password":"a-password","grant_type":"password","client_id":"a-client-id","client_secret":"a-client-secret"}
```
With the bundle `symfony-bundles/json-request-bundle` disabled I get this response :

```json
{
  "error": "unsupported_grant_type",
  "error_description": "The authorization grant type is not supported by the authorization server.",
  "hint": "Check that all required parameters have been provided",
  "message": "The authorization grant type is not supported by the authorization server."
}
```

With the bundle enabled I get this one :
```json
{
  "token_type": "Bearer",
  "expires_in": 2678400,
  "access_token": "an-access-token",
  "refresh_token": "a-refresh-token"
}
```

It's work only with form url encoded request that's why I'm saying it's not working properly 😉

X-Coder264:
That is the expected behavior as per the OAuth 2.0 RFC -> https://tools.ietf.org/html/rfc6749#section-4.1.3

> The client makes a request to the token endpoint by sending the
   following parameters **using the "application/x-www-form-urlencoded"
   format** per Appendix B with a character encoding of UTF-8 in the HTTP
   request entity-body

>  For example, the client makes the following HTTP request using TLS
   (with extra line breaks for display purposes only):

     POST /token HTTP/1.1
     Host: server.example.com
     Authorization: Basic czZCaGRSa3F0MzpnWDFmQmF0M2JW
     Content-Type: application/x-www-form-urlencoded

     grant_type=authorization_code&code=SplxlOBeZQQYbYS6WxSbIA
     &redirect_uri=https%3A%2F%2Fclient%2Eexample%2Ecom%2Fcb

I wouldn't document anything that deviates from the RFC.

vatoer:
base on 
[https://oauth2.thephpleague.com/authorization-server/auth-code-grant/](https://oauth2.thephpleague.com/authorization-server/auth-code-grant/)
```
curl --location --request POST 'https://127.0.0.1:8000/token?response_type=code&client_id=a886e140623cf546e227aff74401b50f&scope=read'
```

trikoder_oauth2.yaml
```
trikoder_oauth2:
    authorization_server:
    # omitted line
        grant_types:
            authorization_code:
                enable:  true
    # omitted line                
```

stil got error 
```
{
    "error": "unsupported_grant_type",
    "error_description": "The authorization grant type is not supported by the authorization server.",
    "hint": "Check that all required parameters have been provided",
    "message": "The authorization grant type is not supported by the authorization server."
}
```

Orkin:
@vatoer you are using query params not post params that you have to put in the body’s request

PR #221, opened 2020-07-31
Add abstract class for user resolve event

Orkin:
If we want to use other grant and get a base structure for the event to retrieve the user


HypeMC:
@Orkin Thank you for your contribution.

PR #223, opened 2020-08-03
Add passing previous exception

Orkin:
Usefull for ConvertExceptionToResponseListener to keep the OAuth2 rfc response format and get the possibility to response a valid json.

I didn't update the listener because it's a BC break. For exemple it can look like this :

```php
<?php

declare(strict_types=1);

namespace Trikoder\Bundle\OAuth2Bundle\EventListener;

use Psr\Http\Message\ResponseFactoryInterface;
use Symfony\Bridge\PsrHttpMessage\Factory\HttpFoundationFactory;
use Symfony\Component\HttpFoundation\Response;
use Symfony\Component\HttpKernel\Event\ExceptionEvent;
use Trikoder\Bundle\OAuth2Bundle\Security\Exception\InsufficientScopesException;
use Trikoder\Bundle\OAuth2Bundle\Security\Exception\OAuth2AuthenticationFailedException;

/**
 * @author Tobias Nyholm <tobias.nyholm@gmail.com>
 */
final class ConvertExceptionToResponseListener
{
    private ResponseFactoryInterface $responseFactory;

    public function __construct(ResponseFactoryInterface $responseFactory) {
        $this->responseFactory = $responseFactory;
    }

    public function onKernelException(ExceptionEvent $event): void
    {
        $exception = $event->getThrowable();
        if ($exception instanceof InsufficientScopesException || $exception instanceof OAuth2AuthenticationFailedException) {
            $leagueException = $exception->getPrevious()->getPrevious(); // not so sexy
            $httpFoundationFactory = new HttpFoundationFactory();
            $event->setResponse($httpFoundationFactory->createResponse($leagueException->generateHttpResponse($this->responseFactory->createResponse())));
        }
    }
}
```

### DIFF
--- a/Security/Firewall/OAuth2Listener.php
+++ b/Security/Firewall/OAuth2Listener.php
@@ -68,7 +68,7 @@ final class OAuth2Listener
             /** @var OAuth2Token $authenticatedToken */
             $authenticatedToken = $this->authenticationManager->authenticate($this->oauth2TokenFactory->createOAuth2Token($request, null, $this->providerKey));
         } catch (AuthenticationException $e) {
-            throw Oauth2AuthenticationFailedException::create($e->getMessage());
+            throw new Oauth2AuthenticationFailedException($e->getMessage(), 401, $e);
         }
 
         if (!$this->isAccessToRouteGranted($event->getRequest(), $authenticatedToken)) {


Stazis555:
Can someone merge this, we need it in our project

Orkin:
@Stazis555 you can add this in your `composer.json`

```
"repositories": [
        {
            "type": "vcs",
            "url": "git@github.com:Orkin/oauth2-bundle.git"
        }
    ],
```

And add this package version : `"trikoder/oauth2-bundle": "v3.1.2"` in your dependencies

It works for me, I'm still waiting the merge too

Security/Firewall/OAuth2Listener.php
```diff
@@ -68,7 +68,7 @@ public function __invoke(RequestEvent $event)
             /** @var OAuth2Token $authenticatedToken */
             $authenticatedToken = $this->authenticationManager->authenticate($this->oauth2TokenFactory->createOAuth2Token($request, null, $this->providerKey));
         } catch (AuthenticationException $e) {
-            throw Oauth2AuthenticationFailedException::create($e->getMessage());
+            throw new Oauth2AuthenticationFailedException($e->getMessage(), 401, $e);
```
X-Coder264, line 71: @Orkin please reuse the current create method. add a `?Throwable $previous = null` parameter to it.

HypeMC: @X-Coder264 Changing the signature of the `change` method would be a breaking change.

X-Coder264: Strictly technically speaking yes it is, but the chance of somebody overriding it is very slim IMHO. Ok, I'll merge this as is then.

btw, it's `create`, not `change` :smile:

PR #230, opened 2020-10-04
Add Symfony 5.1 to CI

HypeMC:


X-Coder264:
As long as the composer.json constraint supports 5.0 we must test it in CI too and unless supporting 5.0 creates an additional maintenance burden for us I wouldn't remove it because from our POV there is no reason to do that.

Issue #38, opened 2019-04-01
Hash client secrets


cGuille:
Hello,

As far as I can see, client secrets are stored in clear text in the database when using the Doctrine persistence.

Storing hashed credentials instead of clear text is a standard security practice. Is there any particular reason not to hash them?

dionsnoeijen:
Is there any progress on this one? I kind of need it. I integrated this library, but found out it's missing this essential security feature. I'm willing to help.

dionsnoeijen:
I have added unit tests to the PR of @Orkin to his PR. Is there any progress on this? https://github.com/Orkin/oauth2-bundle/pull/2

PR #232, opened 2020-10-08
Crypt client secret

Orkin:
Beginning proposal for #38 

For now it's just something really simple but we can imagine add thoses features :
- UpdateSecretCommand to migrate already existing client
- Add field to client model to know if the secret is plain text or crypt to handle secret verification (and to optimise the UpdateSecretCommand, not crypt a secret already crypted)
- Add the ability to "hot crypt" secret => when the `crypt_client_secret` is set to `true` migrate automatically a client when we retrieve it from the database
- Add a new class like `ClientSecretVerifier` to crypt and verify it in one entry point

### DIFF
--- a/Command/CreateClientCommand.php
+++ b/Command/CreateClientCommand.php
@@ -26,11 +26,17 @@ final class CreateClientCommand extends Command
      */
     private $clientManager;
 
-    public function __construct(ClientManagerInterface $clientManager)
+    /**
+     * @var bool
+     */
+    private $cryptClientSecret;
+
+    public function __construct(ClientManagerInterface $clientManager, bool $cryptClientSecret = false)
     {
         parent::__construct();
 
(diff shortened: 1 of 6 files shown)


dionsnoeijen:
Can I help with this one, it should be merged. This seems pretty essential to me 👍 :)

X-Coder264:
@Orkin Instead of having `bool $cryptClientSecret` it'd be better if a Symfony password encoder interface implementation was injected, the plain text encoder by default (because it does nothing) and the user could specify the password encoder implementation which they want via the bundle config. That'd be more flexible than this approach.

Orkin:
5 months after I have found an other solution 😉

X-Coder264:
I'm closing it then. Feel free to reopen the PR if you want to work on the changes. Otherwise we'll work on this when we get the time.

PR #242, opened 2020-11-05
Add a Guard authenticator implementation

### LINKED ISSUE #144: Add a guard implementation
Hi,

I am currently integrating the bundle into a brand new symfony 4.4 project. I wish to protect a certain route with two authentication methods, being JWT Token and OAuth2 Token, depending on the requesting party.
For the JWT Token Auth I use package 'lexik/jwt-authentication-bundle`, which ships with an integrated Security Guard implementation that makes it very easy to add to a firewall section. Your package does not, but I think it would be a great enhancement.

ajgarlag:
Fix #144 

Most of the code borrowed from @yceruto gist at https://gist.github.com/yceruto/f1500b350065481568a3b2971fde8d0f

### DIFF
--- a/Security/Guard/Authenticator/OAuth2Authenticator.php
+++ b/Security/Guard/Authenticator/OAuth2Authenticator.php
@@ -6,12 +6,9 @@ namespace Trikoder\Bundle\OAuth2Bundle\Security\Guard\Authenticator;
 
 use League\OAuth2\Server\Exception\OAuthServerException;
 use League\OAuth2\Server\ResourceServer;
-use Symfony\Bridge\PsrHttpMessage\HttpFoundationFactoryInterface;
 use Symfony\Bridge\PsrHttpMessage\HttpMessageFactoryInterface;
-use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 use Symfony\Component\Security\Core\Exception\AuthenticationException;
@@ -21,7 +18,6 @@ use Symfony\Component\Security\Guard\AuthenticatorInterface;
 use Trikoder\Bundle\OAuth2Bundle\Security\Authentication\Token\OAuth2Token;
 use Trikoder\Bundle\OAuth2Bundle\Security\Authentication\Token\OAuth2TokenFactory;
 use Trikoder\Bundle\OAuth2Bundle\Security\Exception\InsufficientScopesException;
-use Trikoder\Bundle\OAuth2Bundle\Security\Exception\Oauth2AuthenticationFailedException;
 use Trikoder\Bundle\OAuth2Bundle\Security\User\NullUser;
 
 /**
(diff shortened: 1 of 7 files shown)


yceruto:
Thanks for this PR! I will check it this weekend.

yceruto:
Will be good if you change the PR status to make Travis run.

Security/Guard/Authenticator/OAuth2Authenticator.php
```diff
@@ -0,0 +1,143 @@
...
+
+    public function start(Request $request, ?AuthenticationException $authException = null): Response
+    {
+        if ($authException instanceof InsufficientScopesException || $authException instanceof Oauth2AuthenticationFailedException) {
+            $httpException = new HttpException($authException->getCode(), $authException->getMessage(), $authException);
+        } else {
+            $httpException = new UnauthorizedHttpException('Bearer', $authException ? $authException->getMessage() : null, $authException);
+        }
+
+        return new JsonResponse(['code' => $httpException->getCode(), 'message' => $httpException->getMessage()], $httpException->getStatusCode(), $httpException->getHeaders());
```
ajgarlag, line 55: @yceruto I'm not sure why you need a JsonResponse here. I think that the response type should depend on the request format? WDYT?

yceruto: Yes, it's only my end. For general purpose we have to take "the request format" into account.

In Symfony, we use the `ErrorHandler/ErrorRenderer/SerializerErrorRenderer` (introduced in 4.4) to return the right error response according to the request format. Depending on the bundle's min requirements, we could use it.

ajgarlag: I've tested it and is doable, but it introduces a new dependency (`symfony/serializer`) and makes the guard authentication behaviour different from the old firewall.

I think that guard authenticator should mimic the firewall behaviour. Maybe we should introduce `ErrorHandler/ErrorRenderer/SerializerErrorRenderer` in a different PR

froozeify:
Hey,

Firstly I wanted to thanks you about your work on that PR  :-)
And if you need some help on it I can offer some assistance.

While reading what you did and comparing to the Authenticator from the package `lexik/jwt-authentication-bundle`.
I found something very useful in it, and wanted to know if it could be possible to add it also in that PR.
Or, if you prefer, once this one is merged, I can create my own PR implementing it ;-)

The cool stuff they implemented is [**Event** inside the authenticator](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Security/Guard/JWTTokenAuthenticator.php#L244)

What do you think about that ? Could that be done inside this PR, or do you prefer that I add it later when this one is closed ?

Security/Guard/Authenticator/OAuth2Authenticator.php
```diff
@@ -0,0 +1,126 @@
...
+        $oauth2Token->setAuthenticated(true);
+
+        return $oauth2Token;
+    }
+
+    public function onAuthenticationFailure(Request $request, AuthenticationException $exception): ?Response
+    {
+        $this->psr7Request = null;
+
+        throw $exception;
```
yceruto, line 95: Yes, as @froozeify said here https://github.com/trikoder/oauth2-bundle/pull/242#issuecomment-726839920 would be good to implement an extension point to change the response and do what is done here https://github.com/lexik/LexikJWTAuthenticationBundle/blob/2c3448975c5560f4a29ac6f3bfc1fb24673609c0/Security/Guard/JWTTokenAuthenticator.php#L173 (Returning a JsonResponse by default).

The same applies to the `start()` response.

What do you think?

froozeify: I have some more idea, maybe it will be better to wait for this to be merged, and I'll make a merge request after ?

yceruto: Feel free to send the alternative proposal :)

ajgarlag: There are 3 ways to define a security layer in Symfony: Firewall, Guard Authenticator, and HTTP Authenticator (introduced in SF 5.1).

Introducing extension points using Events it's a good thing, but I think it should be added to every security layer implementation, so I'd prefer to see a different PR with your ideas for the Firewall implementation to see how it works.

I think that this PR should mimic Firewall behavior, so if the Event is introduced before this PR gets merged, I'll add it here.

froozeify: That works for me :)   
I'll try to create the PR this week, nice catch for the Firewall, it should totally have the Events too


@yceruto The other alternative (I will probably add it to a separated PR, didn't check but maybe somebody already asked for it) is to add a setting to configure a Response formatter.  
Currently, when an error occurred the response return is not formatted in any type. Just the message in the response.   
For example for my use case, it will be better that the response is in a JSON format (like lexik/jwt do, so my client can catch the error message).   
I was thinking of a setting where we could define a custom Response format (like `none`, `json` or a user custom response class).

For the moment it's just an idea, but a not a good one, like to know some opinion about it

samzijlmans:
@ajgarlag Is there a eta for this merge request? Is there anything you need help with, I would be glad to help

ajgarlag:
@samzijlmans Thanks for your interest. I'm waiting for maintainers review.

X-Coder264 (approved):

X-Coder264:
@ajgarlag @yceruto Thank you for the effort that you've put into this PR.

@froozeify We are interested in adding extension points to the new Guard authenticator and to the "old" system so please feel free to send a PR now that this has been merged.

froozeify:
No problem, I'm currently working on it. And I hope tomorrow a big part will be done :-)

For the moment I didn't create here the PR since it's still in a WIP state, but maybe I should create the PR here ? (first time working on a big PR) I'll take any advice about it :-)

PR #248, opened 2020-12-01
Add PHP 8 and Symfony 5.2 to CI

X-Coder264:


X-Coder264:
@HypeMC I thought this was gonna be a really quick thing to do, but oh boy... :smile: 

For PHP 8 I had to bump Alpine image from Alpine 3.10 to 3.12 since 3.10 is [no longer supported](https://github.com/docker-library/docs/blob/master/php/README.md#supported-tags-and-respective-dockerfile-links) and Xdebug to v3 which supports PHP 8. Unfortunately since Timecop is no longer maintained and we are using it for our tests it has finally bitten us in the ass (as expected) as the PHP 8 image cannot be built because of it as there is no PHP 8 Timecop compatible version released.

As for Symfony 5.2 there are some tests failures which need to be investigated as well as other (random?) failures on the other PHP/Symfony pipelines.

I don't have the time currently to work on this so if anybody wants to contribute please feel free to send a PR for this.

Orkin:
Hello, I worked on a PR for this : #271

PR #250, opened 2020-12-06
Add event in Firewall/Guard

froozeify:
This PR aim to add support of Event inside the Firewall and Guard authenticator.

More details about the starting discussion [here](https://github.com/trikoder/oauth2-bundle/pull/242#issuecomment-726839920)


froozeify:
Wrong branch... see #251

PR #251, opened 2020-12-06
Add event in firewall / guard

froozeify:
This PR aim to add support of Event inside the Firewall and Guard authenticator.

More details about the starting discussion [here](https://github.com/trikoder/oauth2-bundle/pull/242#issuecomment-726839920)


froozeify:
This PR is ready to be merge, for the moment I've created a `response_formatter` but I think the main goal would be to remove it and only use `JsonResponse` for a better consistency.  (maybe in for the next major release ??? )

I've created this formatter to avoid break current user implementation on error handling (currently the error is return directly in the body with no formatting).

X-Coder264:
@froozeify I don't have the time atm to fully review this PR. But just taking a quick look I can't help but notice the removal of the `exception_event_listener_priority` config option and the exception listener. That is a big breaking change or am I missing something? The way I see these events as extension points is that they are being dispatched and if no response is being set on them the code fallbacks to the same behavior of what the current v3.x code does (throwing an exception and having an exception listener handle it). Otherwise I don't see how this could land on a minor v3.x release.

Also I'd prefer the `response_formatter` stuff to be a separate PR after the events stuff gets merged as this is otherwise quite a big change which makes it hard to review (and the events stuff doesn't need the response formatting stuff in order to be merged).

froozeify:
@X-Coder264 Yes the removal of `exception_event_listener_priority` was one point that makes me doubting.  
But the problem that I'm having here is that the implementation of event and `exception_event_listener_priority` stuff can't be together (so no soft adding one and keeping the other side by side with `@deprecation` warning on the old method).

When I start developing the Event feature I was more thinking to add that in a major update but didn't find a v4.x (I should have asked...). And since it was my first PR on a public project forgot about it... (more used to work on company project with a continuously app evolution, so not respecting the Semantic Versioning don't have a major impact)

What I could propose, so the Event is in the v3.x code is:
- ~~Add the event logic (modify with event having nullable response format)~~ Impossible, otherwise event in v3.x will be meaningless, user will have to add response for every event (very tedious I think).  So won't be used, and adding `@deprecation` in the old method will only spam all user logs...
- ~~Add deprecation notification inside `ConvertExceptionToResponseListener` class~~  Also not a good idea since the deprecation will always be throw (the converter is always called by default)

**tl;dr**: I think the event system is awesome 😉 but it can only be added in the next major release and by removing at the same time the old `exception_event_listener_priority` stuff since both can't live together (unless someone has an idea, but don't find one yet)

--- 

For the `response_formatter` stuff, it was an easy method I had found to make the transition to a more consistent app response format (currently almost all are formatted as `JsonResponse` and only the error are in `Response` format).

For that part, I will remove it from this branch.   
But my main question is that with the new Event (that will be added in v4.x ??) maybe we could also formatted by default as `JsonResponse` the error, and so we don't need that formatter.

If you still want it, I can move that code in a separate issue and PR so that custom formatting could be added in a minor v3.x release. (making some user like me happy, because I'd prefer having my frontend always getting JsonResponse, and not having to guess what the format could be).

But more reasonably, I think **we should not add it anymore.**  
Since we could convert the few response that are *wrongly* formatted as `Response` seamlessly with the new Event logic, making after all the response consistent (always formatted in Json).

PR #263, opened 2021-02-26
Allow custom uris for mobile app like com.example.test123:/uri/redire…

AnthonyMatignonCR:
As the redirect uri can be a url but also a custom uri path as described here https://developers.google.com/identity/protocols/oauth2/native-app#custom-uri-scheme

We can't only rely on the filter_var to check if the URI is valid.

### DIFF
--- a/Tests/Model/RedirectUriTest.php
+++ b/Tests/Model/RedirectUriTest.php
@@ -4,13 +4,12 @@ declare(strict_types=1);
 
 namespace Trikoder\Bundle\OAuth2Bundle\Tests\Model;
 
-use \RuntimeException;
 use PHPUnit\Framework\TestCase;
+use RuntimeException;
 use Trikoder\Bundle\OAuth2Bundle\Model\RedirectUri;
 
 class RedirectUriTest extends TestCase
 {
-
     public function testUris()
     {
         $passingUris = [
--- a/Tests/Model/RedirectUriTest.php
+++ b/Tests/Model/RedirectUriTest.php
@@ -17,19 +17,19 @@ class RedirectUriTest extends TestCase
             'http://www.oauth2.com/redirect/to-uri',
             'https://oauth2.com/redirect-to-uri',
             'com.example.test123:/redirect/to-uri',
-            'com.example.test:/redirect-to-uri'
+            'com.example.test:/redirect-to-uri',
         ];
 
         $faillingUris = [
             'someRandomString',
         ];
 
-        foreach($passingUris as $uri) {
+        foreach ($passingUris as $uri) {
             $redirectUri = new RedirectUri($uri);
             $this->assertInstanceOf(RedirectUri::class, $redirectUri);
         }
 
-        foreach($faillingUris as $uri) {
+        foreach ($faillingUris as $uri) {
             $this->expectException(RuntimeException::class);
             $redirectUri = new RedirectUri($uri);
         }
--- a/Tests/Model/RedirectUriTest.php
+++ b/Tests/Model/RedirectUriTest.php
@@ -10,27 +10,44 @@ use Trikoder\Bundle\OAuth2Bundle\Model\RedirectUri;
 
 class RedirectUriTest extends TestCase
 {
-    public function testUris()
+    public function testPassingUris()
     {
         $passingUris = [
             'http://www.oauth2.com/redirect/to-uri',
             'https://oauth2.com/redirect-to-uri',
             'com.example.test123:/redirect/to-uri',
             'com.example.test:/redirect-to-uri',
-        ];
-
-        $faillingUris = [
-            'someRandomString',
+            'com.example.test:',
+            'http://www.oauth2.com',
         ];
 
         foreach ($passingUris as $uri) {
             $redirectUri = new RedirectUri($uri);
             $this->assertInstanceOf(RedirectUri::class, $redirectUri);
         }
+    }
 
-        foreach ($faillingUris as $uri) {
-            $this->expectException(RuntimeException::class);
-            $redirectUri = new RedirectUri($uri);
-        }
+    public function testFailingUri1()
+    {
+        $this->expectException(RuntimeException::class);
+        $redirectUri = new RedirectUri('comexampletest:/redirect/to-uri');
+    }
+
+    public function testFailingUri2()
+    {
+        $this->expectException(RuntimeException::class);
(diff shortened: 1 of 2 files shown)


X-Coder264:
@AnthonyMatignonCR You are absolutely correct, I had this task on my TODO list as one of my projects will need this, but how did you come up with this specific regex?

AnthonyMatignonCR:
I took this base https://www.daniweb.com/programming/web-development/threads/492882/preg-match-to-validate-url-without-http but addedd an extra "?" after the "https?|ftp" section to allow custom uris. 

If you need it, I could add some specific test on the construct RedirectUri against multiple type of uris allowed

X-Coder264:
Yes, please add some unit tests for that so that we know that it works as expected.

AnthonyMatignonCR:
@X-Coder264 I added test and updated Regex as it did pass a random string. It should be better now

AnthonyMatignonCR:
@X-Coder264 Everything is good for the tests.

ajgarlag:
Are we confident enough with the new regular expression?

Maybe, we could use a modified version of Symfony UrlValidator regular expression [pattern](https://github.com/symfony/validator/blob/4.4/Constraints/UrlValidator.php#L25) which is battle-tested. We cannot use it directly because it requires the `host` component of the URL.

Or maybe we can rely on the `parse_url` function to test if all required components are present.

AnthonyMatignonCR:
@ajgarlag I tried different type of string for the validation, but indeed this regular expression is not proof tested enough to validate only the schema of the url. 

I thought about the UrlValidator, but this would require an extra dependency in the composer and it doesn't fit the requirements. 

I didn't think about the `parse_url` method, maybe if we can check only path and scheme with it would be enough. I give it a try.

AnthonyMatignonCR:
@ajgarlag  @X-Coder264 
I tested with a `parse_url` we must validate which keys must be present.

```
$passingUris = [
    'http://www.oauth2.com/redirect/to-uri', //keys : scheme,host,path
    'https://oauth2.com/redirect-to-uri', //keys: scheme,host,path
    'com.example.test123:/', //keys: scheme,path
    'com.example.test:/redirect-to-uri', //keys: scheme,path
];
$faillingUris = [
    'someRandomString', //keys: path
];
```

With this solution we could check if we have scheme and path keys in the parse_url result.

What do you think about it ? Should I switch to parse_url or keep the regexp ?

X-Coder264:
If we can cover all needed cases with `parse_url` I'm all for that as it seems a lot more readable than the regex approach.

ajgarlag:
I would combine current validation with a new one for Custom URI schemes. Something like:
```php
    private function assertValidRedirectUri(string $redirectUri): void
    {
        if (filter_var($redirectUri, FILTER_VALIDATE_URL)) {
            return;
        }

        if ($this->isValidCustomURIScheme($redirectUri)) {
             return;
        }

        throw new RuntimeException(sprintf('The \'%s\' string is not a valid URI.', $redirectUri));
    }

    /**
     * @see https://developers.google.com/identity/protocols/oauth2/native-app?hl=en
     */
    private function isValidCustomURIScheme(string $redirectUri): bool
    {
        $parts = parse_url($redirectUri);
        if (isset($parts['host'])) {
            return false;
        }
        if (!isset($parts['scheme'])) {
            return false;
        }
        if (false === strpos(substr($parts['scheme'], 1 , -1), '.')) {
            return false;
        }
        return true;
    }
```

AnthonyMatignonCR:
To finish this PR, you want to use your solution @ajgarlag ? 
I can update the PR to use your example.

ajgarlag:
@X-Coder264 What do you think?

X-Coder264:
@ajgarlag There is a lot of code branching so it'd be good to add some comments above each path with an explanation of what it's testing/checking because some of them are not obvious at first (and of course we'd need tests for that too).

AnthonyMatignonCR:
I need this update as we release our app soon. 
Could you please tell me when it will be possible to merge it ? 
I will use my fork for the moment to have this feature

ajgarlag:
@AnthonyMatignonCR Can you update the PR to use an `isValidCustomURIScheme` private method similar to the one I proposed in https://github.com/trikoder/oauth2-bundle/pull/263#issuecomment-788686748

It was a quick prototype so I'm sure the code can be improved.

AnthonyMatignonCR:
@ajgarlag It is updated with your suggestions. Tests are all green.

Tests/Model/RedirectUriTest.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        $passingUris = [
+            'http://www.oauth2.com/redirect/to-uri',
+            'https://oauth2.com/redirect-to-uri',
+            'com.example.test123:/redirect/to-uri',
+            'com.example.test:/redirect-to-uri',
+        ];
+
+        $faillingUris = [
+            'someRandomString',
```
ajgarlag, line 23: We should test mor failing URIs
```suggestion
            'someRandomString',
            'comexampletest:/redirect/to-uri',
            '.com.example.test:/redirect/to-uri',
            'com.example.test.:/redirect/to-uri',
            'com.example.test://redirect/to-uri',
```

Tests/Model/RedirectUriTest.php
```diff
@@ -0,0 +1,36 @@
...
+        ];
+
+        foreach ($passingUris as $uri) {
+            $redirectUri = new RedirectUri($uri);
+            $this->assertInstanceOf(RedirectUri::class, $redirectUri);
+        }
+
+        foreach ($faillingUris as $uri) {
```
ajgarlag, line 31:
```suggestion
        foreach ($failingUris as $uri) {
```

ajgarlag (commented):
We should test edge cases

AnthonyMatignonCR:
I added failing tests as independant test, otherwise phpunit won't test other failing tests.
Updated the tests with your new uris suggestions

AnthonyMatignonCR:
I added test about path without URIs, it works too

ajgarlag:
@AnthonyMatignonCR the code looks good to me.

I think you should squash your commits and wait for @X-Coder264 review.

ajgarlag (approved):

HypeMC (approved):

X-Coder264 (approved):

X-Coder264:
Thank you @AnthonyMatignonCR @ajgarlag.

PR #266, opened 2021-03-04
Fix deprecation message on config files

tdutrion:
When running the following command:
```
php bin/console debug:config trikoder_oauth2
```

I get the following deprecations:

```
[info] User Deprecated: Since symfony/dependency-injection 5.1: Not setting the attribute "package" of the node "deprecated" in "/var/www/project/vendor/trikoder/oauth2-bundle/DependencyInjection/../Resources/config/services.xml" is deprecated.
[info] User Deprecated: Since symfony/dependency-injection 5.1: Not setting the attribute "version" of the node "deprecated" in "/var/www/project/vendor/trikoder/oauth2-bundle/DependencyInjection/../Resources/config/services.xml" is deprecated.
```

Adding these attributes fixes it, not sure about the package attribute though :thinking:

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -11,7 +11,7 @@
             <argument type="service" id="Trikoder\Bundle\OAuth2Bundle\Manager\ClientManagerInterface" />
         </service>
         <service id="trikoder.oauth2.league.repository.client_repository" alias="Trikoder\Bundle\OAuth2Bundle\League\Repository\ClientRepository">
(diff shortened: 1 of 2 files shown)


X-Coder264:
@tdutrion Thanks for the PR, but it broke the bundle for <= Symfony 5.0 so we can't merge these changes until we drop Symfony 4.4 and 5.0 support.

tdutrion:
It's only deprecation warnings so no worries :)

I have set the PR as draft to be more explicit, I would assume it would be something for a 4.x branch (dropping a symfony version is kind of a major thing).

PR #270, opened 2021-04-05
Replace ext timecop with Carbon

Orkin:
Change to prepare support for php 8 #248 

I had add a delta for token expiration checking because this is not really the purpose of integration/acceptance test to check that but the purpose of unit test. This is "mandatory" because `league/oauth2-server` use `DateTimeImmutable` object instead of a clock pattern.

I think the delta is the best way to handle that because in production case it can be something that can append so not really a point.

### DIFF
--- a/League/Entity/Client.php
+++ b/League/Entity/Client.php
@@ -10,8 +10,8 @@ use League\OAuth2\Server\Entities\Traits\EntityTrait;
 
 final class Client implements ClientEntityInterface
 {
-    use EntityTrait;
     use ClientTrait;
+    use EntityTrait;
 
     /**
      * @var bool
--- a/Manager/Doctrine/AccessTokenManager.php
+++ b/Manager/Doctrine/AccessTokenManager.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace Trikoder\Bundle\OAuth2Bundle\Manager\Doctrine;
 
-use DateTimeImmutable;
+use Cake\Chronos\Chronos;
 use Doctrine\ORM\EntityManagerInterface;
 use Trikoder\Bundle\OAuth2Bundle\Manager\AccessTokenManagerInterface;
 use Trikoder\Bundle\OAuth2Bundle\Model\AccessToken;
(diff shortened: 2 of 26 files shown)


Orkin:
@X-Coder264 @HypeMC ?

X-Coder264:
@Orkin I'd prefer Carbon over Chronos.

1. Carbon is a lot better maintained (it gets one or two releases per month vs Chronos which gets a couple of releases per year)
2. Carbon is a much more popular choice as **the** PHP datetime library (169 717 086 downloads vs 16 699 937 downloads for Chronos, this is an order of magnitude difference). This means that it's much more likely that projects already require Carbon (as opposed to Chronos) meaning there's a higher chance that we won't be adding a "new" dependency for those users.
3. Carbon has much better CI testing (they test all supported PHP minor versions, but also all PHP patch versions where breaking changes were made to the Datetime PHP API under the hood).
4. Carbon has native Doctrine DBAL type support for Symfony (https://github.com/briannesbitt/Carbon/tree/2.46.0/src/Carbon/Doctrine) which is one of the reasons I always use it over Chronos in my Symfony projects (and it's again one more reason why Symfony users are more likely to use it in their projects over Chronos too).
5. I'm using Carbon in all my Symfony and Laravel applications so I'm generally more comfortable with that choice even though I see that Chronos has basically the same API.

@Orkin @HypeMC Thoughts?

Orkin:
Ok @X-Coder264 I made the change from Chronos to Carbon :)

Orkin:
@X-Coder264 @HypeMC update done 😄

X-Coder264:
@Orkin How did the entire `disable access token saving` feature end up on this branch? :smile: 

Also some tests are missing the call to `CarbonImmutable::setTestNow(null);` in the teardown method.

Orkin:
@X-Coder264 don't worry I will revert, it's just a fail 😅

Orkin:
@X-Coder264 @HypeMC any news about this PR ?

Orkin:
@X-Coder264 @HypeMC a plan to merge this ?

PR #271, opened 2021-04-05
Add support of php 8 and symfony 5.2

Orkin:
Need to merge this PR before #270

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -13,8 +13,8 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: [ '7.2', '7.3', '7.4' ]
-                symfony: [ '4.4', '5.0', '5.1' ]
+                php: [ '7.2', '7.3', '7.4', '8.0' ]
+                symfony: [ '4.4', '5.0', '5.1', '5.2' ]
                 composer-flags: [ '', '--prefer-lowest' ]
                 psr-http-provider: [ 'nyholm', 'zendframework' ]
                 exclude:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -13,7 +13,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: [ '7.2', '7.3', '7.4', '8.0' ]
+                php: [ '7.3', '7.4', '8.0' ]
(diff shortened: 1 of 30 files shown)


Orkin:
I had to drop some support of old php versions and/or some old packages versions so I think it will be better to merge that PR in a major release.

X-Coder264:
@Orkin What is the exact reason why we can't support both PHP 7.2 and PHP 8.0 at the same time (and in 3.x so that we don't have to wait for a major version release just to support PHP 8) ?

As far as I can see Xdebug 3 supports PHP 7.2 and 8 at the same time, the same is true for Carbon/Chronos. As far as Doctrine goes PHP 7.2 users can use DBAL 2.x for PHP < 8 and DBAL 3.x for PHP 8. Doctrine Bundle 1.12.x+ also supports PHP 8 so I don't see the need for dropping the support for Doctrine Bundle 1.x yet.

Orkin:
Hi @X-Coder264 I was faced with some problems :

The first one is a ClassNotFound Exception => https://github.com/doctrine/DoctrineBundle/issues/1197
The second one is a mapping exception fixed here https://github.com/doctrine/orm/pull/8370 and release here https://github.com/doctrine/orm/releases/tag/2.8.1

For doctrine 1.12 I remember that there is tests is not working with php 8 and `--prefer-lowest` tag

I don't remember why I had to remove php 7.2, I will check and come back with some more details

Orkin:
So this is the reason for dropping doctrine bundle 1.12.X
- https://github.com/Orkin/oauth2-bundle/runs/2279635588 with php 7.3, Symfony 4.4 and `--prefer-lowest` flag doctrine bundle 1.8.0 was installed by composer with doctrine/common 3.0.0 and causing `Error: Class 'Doctrine\Common\Persistence\Mapping\Driver\MappingDriverChain' not found` exception
- With php 7.2 it's the same problem, I will try to revert this drop

Orkin:
@X-Coder264 it's fine for PHP 7.2 so if you have any advices for doctrine 1.X support I'll appreciate, I'm not so familiarized with the support of multiple php versions at the same time 😄

Orkin:
@X-Coder264 @HypeMC any news about this PR ?

Orkin:
@X-Coder264 @HypeMC some review here ? Almost 1 month ...

Orkin:
Nice 2y for review

Issue #279, opened 2021-06-02
Doctrine Bundle 2.4 support


vuryss:
Doctrine bundle 2.4 came with some breaking changes, one of which is the default configuration for DBAL connections.

In \Trikoder\Bundle\OAuth2Bundle\DependencyInjection\TrikoderOAuth2Extension::prepend
we have to remove the `'connections' => null,` line - cause it's adding no value, but breaking the compatibility with the following error:

```
  [Symfony\Component\Config\Definition\Exception\InvalidConfigurationException]                                                                                                                                    
  Unrecognized option "connections" under "doctrine.dbal.connections.default". Available options are "MultipleActiveResultSets", "application_name", "auto_commit", "charset", "connectstring", "dbname", "dbname  
  _suffix", "default_dbname", "default_table_options", "driver", "driver_class", "host", "instancename", "keep_replica", "keep_slave", "logging", "mapping_types", "memory", "options", "override_url", "password  
  ", "path", "persistent", "platform_service", "pooled", "port", "profiling", "profiling_collect_backtrace", "profiling_collect_schema_errors", "protocol", "replicas", "schema_filter", "server", "server_versio  
  n", "service", "servicename", "sessionMode", "shard_choser", "shard_choser_service", "shard_manager_class", "shards", "slaves", "sslcert", "sslcrl", "sslkey", "sslmode", "sslrootcert", "unix_socket", "url",   
  "use_savepoints", "user", "wrapper_class".
```

vuryss:
Fixed with doctrine 2.4.2, which reverted their changes in 2.4.1

PR #280, opened 2021-06-02
Update TrikoderOAuth2Extension.php

vuryss:
Fixes https://github.com/trikoder/oauth2-bundle/issues/279

### DIFF
--- a/DependencyInjection/TrikoderOAuth2Extension.php
+++ b/DependencyInjection/TrikoderOAuth2Extension.php
@@ -92,7 +92,6 @@ final class TrikoderOAuth2Extension extends Extension implements PrependExtensio
     {
         $container->prependExtensionConfig('doctrine', [
             'dbal' => [
-                'connections' => null,
                 'types' => [
                     'oauth2_grant' => GrantType::class,
                     'oauth2_redirect_uri' => RedirectUriType::class,


DanielSchmidDS:
Yes. This change fixes my problem in Symfony 5.3.
Thanks.

HypeMC:
@vuryss Unfortunately this would revert https://github.com/trikoder/oauth2-bundle/pull/12 . The solution would need to work on all versions of Doctrine Bundle supported by this bundle without reintroducing old bugs.

vuryss:
@HypeMC which is the bug that #12 solves?

HypeMC:
@vuryss If I remember correctly, without the fix, doctrine bundle would create a "default" connection if one didn't exist which would be a problem if your `default_connection` is named differently.

vuryss:
> @vuryss If I remember correctly, without the fix, doctrine bundle would create a "default" connection if one didn't exist which would be a problem if your `default_connection` is named differently.

This sounds like doctrine bug, not something that should be fixed with this bundle tho...

Anyway do you have a better idea how to fix that and not be stuck with the old bundle version?
I want to start using the attributes that the new doctrine bundle version provides :)

HypeMC:
@vuryss This seems to have been fixed in newer versions https://github.com/doctrine/DoctrineBundle/pull/1356 . I would need to look into it more to give you any concrete answers.

vuryss:
> @vuryss This seems to have been fixed in newer versions [doctrine/DoctrineBundle#1356](https://github.com/doctrine/DoctrineBundle/pull/1356) . I would need to look into it more to give you any concrete answers.

That is the very commit that is causing the oauth2 bundle to break, having that extra connection configuration.

HypeMC:
@vuryss This might get fixed on Doctrine's side https://github.com/doctrine/DoctrineBundle/issues/1366

vuryss:
Yes a lot of discussion is taking place at doctrine bundle side. I guess this is not the only side effect of the last changes.
Linking: 
https://github.com/doctrine/DoctrineBundle/pull/1368
https://github.com/doctrine/DoctrineBundle/issues/1362

I hope it's resolved soon.

vuryss:
I can confirm https://github.com/doctrine/DoctrineBundle/pull/1368 fixes this issue, if it's merged and released I will close this one.
Basically it reverts the problematic https://github.com/doctrine/DoctrineBundle/pull/1356

vuryss:
Fixed in doctrine bundle 2.4.2

PR #281, opened 2021-06-14
Add version 6.0 for the symfony/framework-bundle

Martin1982:
This package is now blocking an upgrade of the Symfony Framework Bundle, add version 6.x to allow upgrading.


X-Coder264:
We can't just allow 6.x as the bundle is not compatible yet. We still have some 5.x deprecations which need to be fixed first (those deprecations can be seen in the CI pipeline) and I'm not sure we can fix all of them without also dropping support for 4.4 in the process (but it should be definitively investigated and fixed at some point).

PR #287, opened 2021-09-02
Fix deprecation notifications since SF 5.1

ricohumme:
symfony/dependency-injection states deprecated tags in configurations should have the package name and the version the deprecation started.
This PR will fix that issue


X-Coder264:
@ricohumme Some builds fail due to some message changes which were recently made in `league/oauth2-server` package. But some builds fail due to this change as IIRC this change is not compatible with Symfony 4.4 and 5.0 which we still support on the `master` branch.

PR #302, opened 2022-05-30
fixing notice error for php 8.1

kuldippujara:
Return type of Trikoder\Bundle\OAuth2Bundle\League\Entity\Scope::jsonSerialize() should either be compatible with JsonSerializable::jsonSerialize(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice.

### DIFF
--- a/League/Entity/Scope.php
+++ b/League/Entity/Scope.php
@@ -14,7 +14,7 @@ final class Scope implements ScopeEntityInterface
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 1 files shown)


X-Coder264:
Fixed by https://github.com/trikoder/oauth2-bundle/pull/308
`mixed` cannot be used as the bundle supports PHP all the way down to PHP 7.2 and mixed is a PHP 8.0 feature

PR #309, opened 2023-06-21
Return NullUser on empty userIdentifier in Oauth2Provider

berislavsapina:
As of symfony 5.4 AuthorizationCheker and Symfony security always expects user to be defined


X-Coder264:
@berislavsapina Closing - this was done via https://github.com/trikoder/oauth2-bundle/pull/311

